PR #1, opened 2016-09-13
First commit (skeleton).

jamalex:


ralphiee22:
Looks good as a starting point. Merging this so I can start to work on the other models.

morango/utils/uuids.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def save(self, *args, **kwargs):
+        
+        if not self.id:
+            self.id = self.calculate_uuid()
+            
+        super(UUIDModelMixin, self).save(*args, **kwargs)
+
+
+class InstanceIDModel(models.Model, UUIDModelMixin):
```
ralphiee22: I think we want UUIDModelMixin to be declared first, so its save method can be found by the MRO.

morango/models.py
```diff
@@ -0,0 +1,85 @@
...
+        # Create the model class itself
+        cls = super(MorangoIndexedModelMeta, mcls).__new__(mcls, name, bases, namespace)
+        
+        # Add a joint index on the index fields to facilitate querying
+        # TODO(jamalex): performance checks to see whether this is the best indexing approach
+        cls._meta.index_together.append(INDEX_FIELDS)
+        
+        return cls
+
+class SerializedModel(models.Model, UUIDModelMixin):
```
ralphiee22: Should also switch the order of inheritance here.

morango/utils/uuids.py
jamalex: Mmm, yeah, that sounds right. Does it work such that calling `super` inside `UUIDModelMixin.save` will properly call Django's `Model.save`, then, too? Because we need to make sure the model still actually gets saved, too. :)

ralphiee22: Shouldn't we just have UUIDModelMixin inherit from models.Model if that's the case? We should also probably set abstract to True for this class, so no tables get created for it.

PR #5, opened 2016-12-20
Several core Morango model related changes.

ralphiee22:
Mostly modifying some fields and basic algorithms for serialization. Moving store and buffer logic to a utilities function which will automatically create the necessary tables for the desired app.


morango/utils/uuids.py
```diff
@@ -3,88 +3,146 @@
...
+    uuid_input_fields = "RANDOM"
+
+    current = models.BooleanField(default=True)
+    date_generated = models.DateTimeField(default=timezone.now)
+    initial_instance_id = models.CharField(max_length=32, blank=True)
+
+
+class InstanceIDModel(UUIDModelMixin):
+
     uuid_input_fields = ("platform", "hostname", "sysversion", "macaddress")
```
jamalex: Also, `database_id`

morango/utils/uuids.py
```diff
@@ -3,88 +3,146 @@
...
+
     platform = models.TextField()
     hostname = models.TextField()
     sysversion = models.TextField()
-    macaddress = models.CharField(blank=True)
-    
+    macaddress = models.CharField(max_length=20, blank=True)
+    database = models.ForeignKey(DatabaseIDModel)
+    counter = models.IntegerField(default=0)
+    current = models.BooleanField(default=True)
```
jamalex: We should also add a `db_path` field, which gets set to the path (on disk) of the current database -- that way, if we have two different databases on the same system, they won't get the same instance ID.

The value for this can be pulled from `settings.DATABASES["default"]["name"]` (which will work even if it's postgres or something else, as it will distinguish between multiple DBs by the DB/connection name). If it's sqlite, and the name doesn't start with a `/`, we should convert it to an absolute path.

requirements/base.txt
```diff
@@ -0,0 +1,7 @@
+pip==8.1.2
+bumpversion==0.5.3
+wheel==0.29.0
+watchdog==0.8.3
+django==1.9.7
+six
```
jamalex: I think there may also be a version of `six` bundled into `django` that you can use instead.

requirements/base.txt
```diff
@@ -0,0 +1,7 @@
...
+django-mptt==0.8.6
```
jamalex: Let's think about how to handle this -- we may not want to make it a base requirement (just for testing), and then have some conditional import logic in morango that is OK if django-mptt doesn't exist. We don't want to force it on a project that won't use it, and we don't want to have version conflicts for projects (like Kolibri) that do use it.

requirements/base.txt
```diff
@@ -0,0 +1,7 @@
+pip==8.1.2
+bumpversion==0.5.3
+wheel==0.29.0
+watchdog==0.8.3
+django==1.9.7
```
jamalex: Rather than pinning, we probably want to do a >= here, to avoid incompatibilities with the dependencies specified in the app (e.g. Kolibri).

requirements/base.txt
```diff
@@ -0,0 +1,7 @@
+pip==8.1.2
+bumpversion==0.5.3
+wheel==0.29.0
+watchdog==0.8.3
```
jamalex: The first 4 deps here are probably implicit, and don't need to be hardcoded in here?

requirements/test.txt
```diff
@@ -0,0 +1,10 @@
+-r base.txt
+# These are for testing
+factory-boy==2.7.0
+fake-factory==0.5.7
+coverage
+mock
+pytest
+mixer
+pytest-cov
+pytest-django
```
jamalex: May be good to pin all of these, to ensure tests are maximally predictable.

morango/utils/uuids.py
```diff
@@ -3,88 +3,146 @@
...
         if not self.id:
             self.id = self.calculate_uuid()
-            
+
         super(UUIDModelMixin, self).save(*args, **kwargs)
 
 
-class InstanceIDModel(models.Model, UUIDModelMixin):
-    
+class DatabaseIDModel(UUIDModelMixin):
```
jamalex: We should add a model manager that overrides `create` to first disable all the other `current` flags. We should do the same in the `save` method, when `id is None`.

jamalex: (And all that should be inside a transaction, so we never have a moment with no current DB ID).

tests/test_uuid_utilities.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class MorangoUUIDUtilsTestCase(TestCase):
+
+    def test_creating_same_instance_ID_model(self):
+        DatabaseIDModel.objects.create()
+        InstanceIDModel.get_or_create_current_instance()
+        firstIDModel = InstanceIDModel.objects.first()
+        (secondIDModel, _) = InstanceIDModel.get_or_create_current_instance()
+        self.assertEqual(firstIDModel, secondIDModel)
```
jamalex: :prepares-for-tests-failing-when-wifi-adapter-is-unplugged-at-just-the-wrong-moment:   :laughing:

jamalex: (or if the kernel is hot-swapped between lines 11 and 13 :rofl: )

tests/test_uuid_utilities.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertEqual(IDModel.macaddress, '')  # assert that macaddress was not added
+
+    def test_uuid_model_mixin(self):
+        # create random object to inherit from UUIDModelMixin
+        childClass = type('ChildClass', (UUIDModelMixin,), {'__module__': 'morango', 'dummy': 1})
+        child = childClass()
+
+        child.uuid_input_fields = 'RANDOM'
+        with mock.patch('uuid.uuid4', return_value='random'):
+            self.assertEqual(child.calculate_uuid(), 'random')
```
jamalex: Faaaaaaaaancy!

tests/test_uuid_utilities.py
```diff
@@ -0,0 +1,46 @@
...
+        with self.assertRaises(AssertionError):
+            child.calculate_uuid()
+
+        child.uuid_input_fields = ()
+        with mock.patch('uuid.uuid4', return_value='random'):
+            self.assertEqual(child.calculate_uuid(), 'random')
+
+        child.uuid_input_fields = ('dummy',)
+        with mock.patch('uuid.uuid5', return_value='DEADBEEF'):
+            self.assertEqual(child.calculate_uuid(), 'DEADBEEF')
```
jamalex: `DEADBEEF` -> `HAPPYCOW`

jamalex: Actually, on this one, might be better to hardcode the expected UUID rather than mock it -- should be predictable, based on knowing what `dummy` was set to above.

morango/models.py
```diff
@@ -1,75 +1,108 @@
...
+    class Meta:
+        abstract = True
+
+    def save(self, set_dirty_bit=True, *args, **kwargs):
+
+        if set_dirty_bit:
+            self._dirty_bit = True
+        super(SyncableModel, self).save(*args, **kwargs)
+
+    def serialize(self, fields=None, exclude=None, *args, **kwargs):
```
ralphiee22: So an example of `_fields_not_to_serialize` would probably be the mptt attributes like `lft`, `rght`, etc.

morango/utils/morango_mptt.py
```diff
@@ -0,0 +1,40 @@
...
+
+    class Meta:
+        abstract = True
+
+    def save(self, set_dirty_bit=True, *args, **kwargs):
+
+        assert isinstance(self, SyncableModel), "Model `{}` should also inherit from `SyncableModel` in morango.".format(self._meta.model_name)
+        # we get calculate_uuid and dirty bit from syncable model
+        if not self.id:
+            self.id = self.calculate_uuid()
```
ralphiee22: yeah currently looking into this. The mro is not working as i thought it was...

morango/utils/uuids.py
```diff
@@ -3,88 +3,146 @@
...
+    """
+    Model to be used for tracking database ids.
+    """
+
+    uuid_input_fields = "RANDOM"
+
+    current = models.BooleanField(default=True)
+    date_generated = models.DateTimeField(default=timezone.now)
+    initial_instance_id = models.CharField(max_length=32, blank=True)
```
ralphiee22: It was just supposed to be metadata about what the current instance id was when the new database id was created. Not entirely sure how helpful that would be here.

morango/models.py
jamalex: Yep! The names of those fields are [configurable](https://django-mptt.github.io/django-mptt/models.html#model-options) through MPTT, so probably better to pull them from the model's options.

jamalex: If it's hard to pull them during `__init__` to set `_fields_not_to_serialize`, you could instead just override `serialize` on `MorangoMPTTModel` to run super and then delete those keys from the dict before returning it.

Actually, that would probably be the best way to do it in this case, as it then still leaves `_fields_not_to_serialize` for use by the subclasses, to exclude additional fields if needed.

morango/utils/morango_mptt.py
jamalex: mroAR!!!! :lion:

morango/models.py
```diff
@@ -1,75 +1,108 @@
...
 ###################################################################################################
-# METADATA: Additional data to keep track of the state of the system (sessions, counters, etc)
+# APP MODELS: Abstract models from which app models should inherit in order to make them syncable
 ###################################################################################################
+class SyncableModelQuerySet(models.query.QuerySet):
 
+    def update(self, **kwargs):
+        kwargs.update({'_dirty_bit': True})
+        super(SyncableModelQuerySet, self).update(**kwargs)
+    update.queryset_only = True  # Unsure whether django will not place this on manager class by default
```
ralphiee22: Agreed. I can't remember why I did it this way but yeah I think it should also be on the manager.

morango/utils/register_models.py
```diff
@@ -0,0 +1,77 @@
...
+        class Meta:
+            app_label = label
+
+    class BaseSyncableModel(SyncableModel):
+        """
+        ``BaseSyncableModel`` is where classes should inherit from if they want to make their data syncable.
+        """
+
+        _morango_profile = profile
+        _morango_partitions = partitions
```
ralphiee22: @jamalex I was thinking that we would need it on the model for the function `get_partition_names`. I always thought of `_profiles` as something that would only be used in morango, but for this case we would need to be able to access it in each model in the app.

morango/models.py
ralphiee22: @jamalex `MorangoMPTTModel` does not inherit from `SyncableModel` though. So I wouldn't really be overriding `serialize`. I would probably have to replicate the logic in `serialize` and remove the keys from the dict then.

morango/utils/register_models.py
jamalex: For `get_partition_names`, it might just return a subset, and it needs to know where it's pulling from etc, so it'll need to hardcode anyway (i.e. won't be looping over a list of possible partitions). So I think it can be left off here, and just stored in the profile cache. The function that calls `get_partition_names` can do validation on the result to make sure it's not including extra stuff.

morango/models.py
```diff
@@ -10,78 +12,72 @@
 ###################################################################################################
 class SyncableModelQuerySet(models.query.QuerySet):
 
-    def update(self, **kwargs):
-        kwargs.update({'_dirty_bit': True})
+    def update(self, dirty_bit_signal=True, **kwargs):
```
jamalex: Since "signal" has a specific meaning in Django, might be better to call it something like `update_dirty_bit_to`?

morango/models.py
```diff
@@ -10,78 +12,72 @@
...
-        return cls(**kwargs).save()
+        return cls(**kwargs)
 
     @classmethod
     def merge_conflict(cls, current, incoming):
         return incoming
 
-    def get_shard_indices(self, *args, **kwargs):
+    def get_partition_names(self, *args, **kwargs):
         """Should return a dictionary with any relevant shard index keys included, along with their values."""
```
jamalex: Shard index -> partition

morango/models.py
```diff
@@ -100,10 +96,33 @@ class AbstractStoreModel(models.Model):
...
+    max_counter = models.IntegerField()
+    transfer_session_id = UUIDField()
+    kind = models.BooleanField()
 
     class Meta:
         abstract = True
 
+    @classmethod
+    def filter_max_counters(cls, query):
+        query = json.loads(query)
```
jamalex: Better to have it be a dict already -- on the side that's creating the filter specification/query, it shouldn't have to first convert to a string before using this function.

morango/models.py
```diff
@@ -100,10 +96,33 @@ class AbstractStoreModel(models.Model):
...
+    instance_id = models.UUIDField()
+    max_counter = models.IntegerField()
+    transfer_session_id = UUIDField()
+    kind = models.BooleanField()
 
     class Meta:
         abstract = True
 
+    @classmethod
+    def filter_max_counters(cls, query):
```
jamalex: A clearer name (so it doesn't look like it's just a queryset-style filter) might be `get_max_counters_for_filter`. And let's call the argument "filter" rather than "query".

morango/models.py
```diff
@@ -100,10 +96,33 @@ class AbstractStoreModel(models.Model):
...
+        query = json.loads(query)
+        filters = []
```
jamalex: Perhaps then call this "queries" rather than "filters".

morango/models.py
```diff
@@ -100,10 +96,33 @@ class AbstractStoreModel(models.Model):
...
+        for key, value in six.iteritems(query):
+            filters.append(Q(**{key: value}) | Q(**{key: "*"}))
+
+            query = reduce(lambda x, y: x & y, filters)
+            rows = cls.objects.filter(query)
+            return rows.values('instance_id').annotate(max_counter=Max('max_counter'))
```
jamalex: Does it not need an `aggregate` or something in here? (I'm never quite sure, and it ends up being a bit of trial and error). Does this behave properly with multiple matching rows for the same instance_id?

ralphiee22: I'm pretty sure its annotate. I wrote some preliminary tests for this and it seemed to be working fine. I can push those tests in a separate PR.

morango/utils/morango_mptt.py
```diff
@@ -1,40 +1,29 @@
...
+class MorangoMPTTTreeManager(managers.TreeManager):
+
+    def get_queryset(self):
+        return MorangoTreeQuerySet(self.model, using=self._db)
+
+    @managers.delegate_manager
+    def _mptt_update(self, qs=None, **items):
+        if qs is None:
+            qs = self
+        return qs.update(dirty_bit_signal=None, **self._translate_lookups(**items))
```
jamalex: Should work to just add the dirty bit argument to the kwargs and call `super`, rather than duplicating the logic here. Won't need to add the decorator here then either.

morango/utils/register_models.py
```diff
@@ -3,16 +3,30 @@
...
-from morango.constants import buffer_kinds, morango_structures
+from django.utils import six
+from morango.models import AbstractStoreModel, AbstractDatabaseMaxCounter, SyncableModel
+
 from .controller import _profiles
 
+Profile = namedtuple('Profile', 'partitions buffer store database_max_counter')
+
+
+def register_morango_profile(profile="", partitions=(), module=""):
```
jamalex: Easiest way to indicate that these aren't optional arguments would be to not have default values here.

morango/utils/uuids.py
```diff
@@ -104,21 +106,39 @@ def save(self, *args, **kwargs):
         super(UUIDModelMixin, self).save(*args, **kwargs)
 
 
+class DatabaseManager(models.Manager):
+
+    def create(self, **kwargs):
+
+        # set current flag to false for all database_id models
+        DatabaseIDModel.objects.update(current=False)
+        super(DatabaseManager, self).create(**kwargs)
```
jamalex: Safest to do these two operations within some sort of transaction, so we never have a limbo state with no "current" database ID.

jamalex: And is `create` meant to return a value?

morango/utils/uuids.py
```diff
@@ -104,21 +106,39 @@ def save(self, *args, **kwargs):
...
 
+    objects = DatabaseManager()
+
     current = models.BooleanField(default=True)
     date_generated = models.DateTimeField(default=timezone.now)
     initial_instance_id = models.CharField(max_length=32, blank=True)
 
+    def save(self, *args, **kwargs):
+
+        if not self.id:
```
jamalex: `and self.current` -- to reduce chances of having none

morango/utils/uuids.py
```diff
@@ -127,6 +147,7 @@ class InstanceIDModel(UUIDModelMixin):
     database = models.ForeignKey(DatabaseIDModel)
     counter = models.IntegerField(default=0)
     current = models.BooleanField(default=True)
+    db_path = models.CharField(max_length=100)
```
jamalex: May want a slightly longer `max_length` (1000?), since we can't control the paths people are putting their DBs in.

morango/models.py
jamalex: Yeah, I think `values` already does a `GROUP BY` when it's combined with `annotate`. :magic:

Tests would be good; can come in another PR, yep.

jamalex:
Looking good! Added some new comments, and also still this older change to address:
https://github.com/learningequality/morango/pull/5/files#r94182715

tests/test_uuid_utilities.py
ralphiee22: sorry, this will come in another PR. I moved these tests inside the testapp.

morango/utils/morango_mptt.py
```diff
@@ -11,11 +11,9 @@ class MorangoMPTTTreeManager(managers.TreeManager):
     def get_queryset(self):
         return MorangoTreeQuerySet(self.model, using=self._db)
 
-    @managers.delegate_manager
     def _mptt_update(self, qs=None, **items):
-        if qs is None:
-            qs = self
-        return qs.update(dirty_bit_signal=None, **self._translate_lookups(**items))
+        items['update_dirty_bit_to'] = None
+        return super(MorangoMPTTTreeManager, self)._mptt_update(qs, **items)
```
ralphiee22: Resolves https://github.com/learningequality/morango/issues/6

morango/utils/uuids.py
```diff
@@ -110,9 +110,12 @@ class DatabaseManager(models.Manager):
 
     def create(self, **kwargs):
 
-        # set current flag to false for all database_id models
-        DatabaseIDModel.objects.update(current=False)
-        super(DatabaseManager, self).create(**kwargs)
+        # do within transaction so we always have a current database ID
+        with transaction.atomic():
```
jamalex: Were you able to confirm that this kind of thing does guarantee atomicity? E.g. by putting pdb breaks in between, and then opening an additional shell in a separate process and connecting to the DB to check? Or using sqlitebrowser?

jamalex:
Awesome! Question on there, but all looks good!

PR #7, opened 2017-01-10
Tests for testapp

ralphiee22:
## Summary

Updated some settings to get tests working with testapp. Created models for testapp and moved uuid utils tests inside it. Testing dirty bit is set correctly, and other various tests.

## TODO


tests/testapp/tests/test_morango_mptt.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertFalse(isinstance(MyUser.objects.all(), MorangoTreeQuerySet))
+
+    def test_mptt_manager_update(self):
+        self.assertTrue(Facility.objects.first()._dirty_bit)
+        Facility.objects.update(update_dirty_bit_to=False)
+        self.assertFalse(Facility.objects.first()._dirty_bit)
+        Facility.objects.update(update_dirty_bit_to=None)
+        self.assertFalse(Facility.objects.first()._dirty_bit)
+        Facility.objects.update(update_dirty_bit_to=True)
+        self.assertTrue(Facility.objects.first()._dirty_bit)
```
jamalex: For complete completeness, perhaps also add this at the end:
```
        Facility.objects.update(update_dirty_bit_to=None)
        self.assertTrue(Facility.objects.first()._dirty_bit)
```

jamalex: (and below)

tests/testapp/tests/test_morango_mptt.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def test_mptt_save(self):
+        fac = Facility.objects.first()
+        self.assertTrue(fac._dirty_bit)
+        fac.save(update_dirty_bit_to=False)
+        self.assertFalse(Facility.objects.first()._dirty_bit)
+        fac.save(update_dirty_bit_to=None)
+        self.assertFalse(Facility.objects.first()._dirty_bit)
+        fac.save(update_dirty_bit_to=True)
+        self.assertTrue(Facility.objects.first()._dirty_bit)
```
jamalex: Maybe this is handled elsewhere, but making sure that a plain save (with no `update_dirty_bit_to` argument provided) also sets the dirty bit would be good.

tests/testapp/tests/test_serialization.py
```diff
@@ -0,0 +1,37 @@
...
+from morango.utils.syncing_utils import _syncing_models
+
+
+class SerializationTestCase(TestCase):
+
+    def setUp(self):
+        self.bob = Facility.objects.create(name="bob")
+        self.dob = Facility.objects.create(name="dob")
+        self.bob.id = self.bob.id.hex
+        self.dob.id = self.dob.id.hex
```
jamalex: Probably good for serialization to be able to handle this case (`id` being a UUID object) internally. I believe that can be achieved by properly definition of certain methods on the `Field` class -- we have other `UUIDField` instances in other places (e.g. Kolibri) that may do that better.

tests/testapp/tests/test_serialization.py
```diff
@@ -0,0 +1,37 @@
...
+from django.test import TestCase
+from facility_profile.models import Facility
+from morango.utils.syncing_utils import _syncing_models
+
+
+class SerializationTestCase(TestCase):
+
+    def setUp(self):
+        self.bob = Facility.objects.create(name="bob")
+        self.dob = Facility.objects.create(name="dob")
```
jamalex: You write the best tests.

tests/testapp/tests/test_serialization.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_serialization(self):
+        self.assertEqual(self.bob_dict['name'], 'bob')
+        self.assertEqual(self.bob.morango_model_name, Facility.morango_model_name)
+
+    def test_field_deserialization(self):
+        class_model = _syncing_models[self.bob.morango_model_name]
+        self.bob_copy = class_model.deserialize(self.bob_dict)
+        for f in Facility._meta.concrete_fields:
+            if isinstance(f, models.DateTimeField):
+                continue
```
jamalex: Why skip DateTimeField?

tests/testapp/tests/test_syncing_models.py
```diff
@@ -0,0 +1,43 @@
...
+        self.assertTrue(isinstance(MyUser.objects.all(), SyncableModelQuerySet))
+
+    def test_syncable_manager_update(self):
+        self.assertTrue(MyUser.objects.first()._dirty_bit)
+        MyUser.objects.update(update_dirty_bit_to=False)
+        self.assertFalse(MyUser.objects.first()._dirty_bit)
+        MyUser.objects.update(update_dirty_bit_to=None)
+        self.assertFalse(MyUser.objects.first()._dirty_bit)
+        MyUser.objects.update(update_dirty_bit_to=True)
+        self.assertTrue(MyUser.objects.first()._dirty_bit)
```
jamalex: As above.

morango/utils/syncing_utils.py
```diff
@@ -0,0 +1,14 @@
+_syncing_models = {}  # dictionary of models with morango naming identifier
+
+
+def add_syncing_models():
+    from morango.models import SyncableModel
+    import django.apps
+
+    model_list = []
+    for model_class in django.apps.apps.get_models():
+        if issubclass(model_class, SyncableModel) and not model_class._meta.proxy:
```
ralphiee22: Abstract subclasses do not show up in `get_models`.

morango/utils/syncing_utils.py
```diff
@@ -0,0 +1,14 @@
...
+            model_list.append(model_class)
+
+    for model_class in model_list:
+        _syncing_models.update({model_class.morango_model_name: model_class})
```
ralphiee22: I've fleshed this out more because now I have a dictionary based on profile. I have also ordered models in those profiles, based on foreign key dependencies.

pytest.ini
```diff
@@ -1,4 +1,4 @@
 [pytest]
 testpaths = tests/ morango/
-DJANGO_SETTINGS_MODULE = morango.settings
+DJANGO_SETTINGS_MODULE = testapp.settings
```
ralphiee22: right they should work in the testapp context. But I am writing all my tests to happen in the testapp.

morango/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import, print_function, unicode_literals
+
+default_app_config = 'morango.apps.MorangoConfig'
```
ralphiee22: yeah our minimal django will be probably be 1.8 or 1.9.
If we put the code there, that will also be called everytime on app load, correct?
If that's the case, then I might prefer having it there.

tests/testapp/facility_profile/models.py
```diff
@@ -0,0 +1,78 @@
...
+    morango_model_name = "contentsummarylog"
+    uuid_input_fields = ("user_id", "content_id")
+
+    user = models.ForeignKey(MyUser)
+    content_id = UUIDField(db_index=True, default=uuid.uuid4)
+
+
+class DummyModel(models.Model):
+
+    morango_model_name = "dummymodel"
```
ralphiee22: I was going to do something with this for a test, but I didn't end up using it.

tests/testapp/facility_profile/models.py
```diff
@@ -0,0 +1,78 @@
...
+            self.kind = self._KIND
+
+
+class ProxyManager(models.Manager):
+    pass
+
+
+class ProxyModel(ProxyParent):
+
+    morango_model_name = 'proxy'
```
ralphiee22: I think we should chat about this tomorrow, in the case of kolibri's proxy models.

tests/testapp/testapp/settings.py
```diff
@@ -78,7 +79,7 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
+        'NAME': os.path.join(BASE_DIR, 'testapp.db'),
```
ralphiee22: I was just giving it the same name as the project.

tests/testapp/tests/test_morango_mptt.py
ralphiee22: yeah forgot that one!

tests/testapp/tests/test_serialization.py
ralphiee22: I should probably change the names before someone with dyslexia reads my code.

tests/testapp/tests/test_serialization.py
ralphiee22: What I realized is that through serialization and deserialization of this field, we lose 3 digits of precision, so they do not come out equal. I could probably rewrite the test to account for this.

PR #8, opened 2017-03-02
Schema refactor and tests

ralphiee22:
## Summary

Adding more models that would be related to syncing. Now have record max counter as its own data structure, as well as its own buffer. Moved tests around and added a few more.

## TODO

## Reviewer guidance

Pay close attention models.py and the fields on the models to make sure I did not miss anything.

### DIFF
--- a/morango/apps.py
+++ b/morango/apps.py
@@ -24,11 +24,8 @@ class MorangoConfig(AppConfig):
             if not DatabaseIDModel.objects.all():
                 DatabaseIDModel.objects.create()
             InstanceIDModel.get_or_create_current_instance()
-        except OperationalError as e:
-            if 'no such table' in e.message:
-                logging.info("Morango models have not been migrated: {}".format(e.message))
-            else:
-                logging.error("OperationalError on application load: {}".format(e.args))
+        except OperationalError:
+            pass
 
         # add models to be synced by profile
         add_syncing_models()
--- a/morango/apps.py
+++ b/morango/apps.py
@@ -1,7 +1,12 @@
 from __future__ import unicode_literals
 
+import logging as logger
+
 from django.apps import AppConfig
+from django.db.utils import OperationalError
 from morango.utils.syncing_utils import add_syncing_models
+
+logging = logger.getLogger(__name__)
 
 
 class MorangoConfig(AppConfig):
@@ -9,4 +14,21 @@ class MorangoConfig(AppConfig):
     verbose_name = 'Morango'
 
     def ready(self):
+        from morango.models import DatabaseIDModel, InstanceIDModel
+
+        # NOTE: Warning: https://docs.djangoproject.com/en/1.10/ref/applications/#django.apps.AppConfig.ready
+        # its recommended not to execute queries in this method, but we are producing the same result after the first call, so its OK
+
+        # call this on app load up to get most recent system config settings
+        try:
+            if not DatabaseIDModel.objects.all():
+                DatabaseIDModel.objects.create()
+            InstanceIDModel.get_or_create_current_instance()
+        except OperationalError as e:
+            if 'no such table' in e.message:
+                logging.info("Morango models have not been migrated: {}".format(e.message))
+            else:
+                logging.error("OperationalError on application load: {}".format(e.args))
+
+        # add models to be synced by profile
         add_syncing_models()
--- a/morango/models.py
+++ b/morango/models.py
@@ -1,37 +1,216 @@
-from django.db import models
+import os
+import platform
+import sys
+import uuid
+
+from django.conf import settings
+from django.db import models, transaction
 from django.db.models import Max, Q
-from django.utils import six
+from django.utils import six, timezone
 from django.utils.encoding import python_2_unicode_compatible
 
+from .manager import SyncableModelManager
 from .utils.uuids import UUIDModelMixin, UUIDField
 
 
-###################################################################################################
-# APP MODELS: Abstract models from which app models should inherit in order to make them syncable
-###################################################################################################
-class SyncableModelQuerySet(models.query.QuerySet):
+class DatabaseManager(models.Manager):
+    """
+    We override ``model.Manager`` in order to wrap creating a new database ID model within a transaction. With the
+    creation of a new database ID model, we set all previously created models current flag to False.
+    """
 
-    def update(self, update_dirty_bit_to=True, **kwargs):
-        if update_dirty_bit_to is None:
-            pass  # don't do anything with the dirty bit
-        elif update_dirty_bit_to:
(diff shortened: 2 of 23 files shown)


morango/utils/register_models.py
```diff
@@ -50,29 +51,34 @@ def __new__(mcls, name, bases, namespace):
...
-        ``StoreModel`` is where serialized data is persisted, along with metadata about counters and history.
-        """
+    class StoreModel(six.with_metaclass(MorangoIndexedModelMeta, AbstractStore)):
+
+        class Meta:
+            app_label = label
+
+    class RecordMaxCounter(AbstractRecordMaxCounter):
+
+        store_model = models.ForeignKey(StoreModel)
```
ralphiee22: should record max counters foreign key onto to the store model, or just have the pk of the store model for this field?

morango/utils/register_models.py
```diff
@@ -50,29 +51,34 @@ def __new__(mcls, name, bases, namespace):
...
+
+        class Meta:
+            app_label = label
+
+    class RecordMaxCounterBuffer(RecordMaxCounter):
```
ralphiee22: Depending on above, should we have rmc buffer foreign key onto the store buffer?

morango/settings.py
```diff
@@ -1,18 +1,22 @@
 # Django settings for morango project.
+import os
```
jamalex: Does this settings file really belong in the morango root, or within the testapp?

morango/utils/register_models.py
jamalex: We had talked about it being an actual foreign key -- but it was made messier by the fact that there were multiple store models (one per profile). With that upcoming refactor, becomes easier.

morango/utils/register_models.py
jamalex: I think for these we decided maybe not, because the PK is auto-generated and so we can't set it in the single-SQL-query transfer. But I'm not sure what the alternative is (I guess storing both the transfer ID and the UUID) -- let's chat about this piece.

morango/utils/syncing_utils.py
```diff
@@ -1,14 +1,93 @@
-_syncing_models = {}  # dictionary of models with morango naming identifier
```
jamalex: Maybe we should rename this file to just "syncing.py", since it's in a utils folder already.

morango/utils/syncing_utils.py
```diff
@@ -1,14 +1,93 @@
...
+                model_class._meta.proxied_children.reverse()
+                for proxy_model in model_class._meta.proxied_children:
+                    _insert_model_into_profile_dict(proxy_model.model, profile)
+            else:
+                _insert_model_into_profile_dict(model_class, profile)
 
-    for model_class in model_list:
-        _syncing_models.update({model_class.morango_model_name: model_class})
+    # for each profile, create a dict mapping from morango model names to model class
+    for profile, model_list in iteritems(_profiles_for_syncing_models):
```
jamalex: The more I read the variable name `_profiles_for_syncing_models`, the less it means to me. Anything more intuitive we can find?

setup.cfg
```diff
@@ -15,4 +15,36 @@ replace = __version__ = '{new_version}'
 universal = 1
 
 [flake8]
-exclude = docs
+ignore = E226,E302,E41
+max-line-length = 160
```
jamalex: Woooooooooooooooooo! So generous! And here I was trying to hold myself back to ~100!

tests/testapp/tests/test_serialization.py
```diff
@@ -1,34 +1,34 @@
...
 from facility_profile.models import Facility
-from morango.utils.syncing_utils import _syncing_models
+from morango.utils.syncing_utils import _profiles_for_syncing_models
 
 
 class SerializationTestCase(TestCase):
 
     def setUp(self):
         self.bob = Facility.objects.create(name="bob")
-        self.dob = Facility.objects.create(name="dob")
```
jamalex: :cry:

jamalex: Dob, we barely knew you...

morango/models.py
```diff
@@ -1,100 +1,155 @@
...
-    id = UUIDField(max_length=32, primary_key=True)
+    id = UUIDField(primary_key=True)
     serialized = models.TextField(blank=True)
     deleted = models.BooleanField(default=False)
     version = models.CharField(max_length=40)
     history = models.TextField(blank=True)
     last_saved_instance = models.UUIDField()
     last_saved_counter = models.IntegerField()
-    last_saved_counter_per_instance = models.TextField(default="{}")  # RMC
+    record_max_counters = models.TextField(default="{}")
```
jamalex: Hmm, you do implement them, below, but this is also still here. Can discuss.

morango/apps.py
```diff
@@ -1,12 +1,31 @@
...
+
+        # NOTE: Warning: https://docs.djangoproject.com/en/1.10/ref/applications/#django.apps.AppConfig.ready
+        # its recommended not to execute queries in this method, but we are producing the same result after the first call, so its OK
+
+        # call this on app load up to get most recent system config settings
+        try:
+            if not DatabaseIDModel.objects.all():
+                DatabaseIDModel.objects.create()
+            InstanceIDModel.get_or_create_current_instance()
+        except OperationalError:
```
ralphiee22: Since the `ready` method in apps.py is called when the app is loaded, and during management commands, if there is no initial database, then the above queries will fail. This was a workaround to allow it to `migrate` a database without making any queries.

morango/models.py
```diff
@@ -1,100 +1,155 @@
...
+        mac = uuid.getnode()
+        if (mac >> 40) % 2 == 0:  # 8th bit (of 48 bits, from left) is 1 if MAC is fake
+            kwargs["macaddress"] = mac
+
+        # do within transaction so we only ever have 1 current instance ID
+        with transaction.atomic():
+            obj, created = InstanceIDModel.objects.get_or_create(**kwargs)
+            if created:
+                InstanceIDModel.objects.exclude(id=obj.id.hex).update(current=False)
+                return obj, created
```
ralphiee22: yes.

morango/models.py
ralphiee22: just forgot to remove it.

morango/settings.py
ralphiee22: yeah this is where we should set all morango related settings.

morango/utils/syncing_utils.py
ralphiee22: what about `_syncable_models_for_profile`?
so it would be like `syncable_models_for_profile['facilitydata']`

PR #9, opened 2017-03-10
Models overhaul and skeleton for controllers.

ralphiee22:
## Summary

Re-worked models to be inside morango as per discussions. Layed out some structure for profile controller and sync controllers. Name field changes. Added signals for deleted models.

## TODO


morango/utils/syncsession.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def _join_with_logical_operator(lst, operator):
+    op = ") {operator} (".format(operator=operator)
+    return "(({items}))".format(items=op.join(lst))
+
+
+class SyncConnection(object):
+
+    def get_remote_certifcates():
```
jamalex: *certificates

morango/utils/syncsession.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, host, profile):
+        self.host = host
+        self.profile = profile
+        self.session_id = uuid.uuid4()
+        self.transfer_session_id = None
+        self.symmetric_key = 'xxx'
+        self.sync_session = SyncSession(host=host, id=self.session_id, symmetric_key=self.symmetric_key)
+
+    def initiate_push_request(self, sync_filter, chunksize):
```
jamalex: Can remove "request"

morango/utils/syncsession.py
```diff
@@ -0,0 +1,111 @@
...
+        pass
+
+    def initiate_pull_request(self, sync_filter, fsics, transfer_id):
```
jamalex: Can remove "request"

morango/utils/syncsession.py
```diff
@@ -0,0 +1,111 @@
...
+
+    @transaction.atomic
+    def _queue_into_buffer(self, fsics):
+        """
+        Takes a chunk of data from the store to be put into the buffer to be sent to another morango instance.
+        """
+        where_statements = []
+        # create condition for all incoming FSICs where instance_ids are equal, but internal counters are higher then FSICs counters
+        for instance, counter in iteritems(fsics):
+            where_statements += ['(last_saved_instance == {0} AND last_saved_counter > {1})'.format(instance, counter)]
```
jamalex: We also need to make sure (doesn't need to be here, necessarily -- can possibly be when we calculate the FSICs) that we include all instances that the other side doesn't even know about -- i.e. the "implicit counter=0" instances.

jamalex: Note that the FSIC coming from the other device can't know about instances it hasn't seen, so it would either need to be updated before calling this function to add the "implicit 0" entries, or we could add a clause along the lines of "OR (instance not in (list,of,known,ids))"

morango/utils/syncsession.py
```diff
@@ -0,0 +1,111 @@
...
+        condition = _join_with_logical_operator(where_statements, 'OR')
+        # execute raw sql to take all records that match condition, to be put into buffer for transfer
+        with connection.cursor() as cursor:
+            queue_buffer = '''INSERT INTO {outgoing}
+                            (serialized, deleted, last_saved_instance, last_saved_counter,
+                             model_name, profile, partitions, transfer_session_id, incoming_buffer)
+                            SELECT serialized, deleted, last_saved_instance, last_saved_counter,
+                            model_name, profile, partitions, {transfer_session_id}, {incoming_buffer}
+                            FROM {store}
+                            [WHERE {condition}]'''.format(outgoing=DataTransferBuffer._meta.db_table,
```
jamalex: We'll also want to filter by profile.

morango/utils/syncsession.py
```diff
@@ -0,0 +1,111 @@
...
+                                           transfer_session_id=self.transfer_session_id,
+                                           record_max_counter=RecordMaxCounter._meta.db_table,
+                                           outgoing_buffer=DataTransferBuffer._meta.db_table)
+            cursor.execute(queue_rmc_buffer)
+
+    @transaction.atomic
+    def _integrate_into_store(self, session):
+        """
+        Takes data from the buffers and merges into the store and record max counters.
+        """
```
jamalex: We'll want to do all the "fast forwards" as a single SQL query block transfer (that can just overwrite the store entries, since it's a fast forward), and then another query to get all the ones that are merge conflicts -- and those we *will* need to loop over individually.

tests/testapp/tests/test_signals.py
```diff
@@ -0,0 +1,29 @@
...
+        InstanceIDModel.get_or_create_current_instance()
+        [FacilityModelFactory() for _ in range(10)]
+        self.mc = MorangoProfileController('facilitydata')
+        self.mc._serialize_into_store()
+
+    def test_deleted_flag_gets_set(self):
+        facility = Facility.objects.first()
+        deleted_id = facility.id
+        facility.delete()
+        self.assertTrue(DeletedModels.objects.filter(id=deleted_id))
```
jamalex: Would be good to also add an explicit test case for CASCADE deletes, as I'm not actually sure how that interacts with signals (may be at a pure DB level). So, if you create a facility, then create a user in it, then delete the facility, is the user's ID in the `DeletedModels` table?

jamalex: Ah, looks like we'll need to handle this manually: http://stackoverflow.com/questions/27566614/django-cascade-delete-and-post-delete-signal

(override the delete methods, and scan through models that are going to be cascade deleted, so we can make sure to mark their IDs in the deleted models table)

ralphiee22: just tested this. the signal still gets called on cascading delete

PR #11, opened 2017-05-09
Change the way we calculate uuids.

ralphiee22:
## Summary

We now calculate uuids based on the partition values of the model, as well as the source id (which will be unique fields on the model or a random uuid). 

NOTE:
Tests will break, but we will fix at a later time.

### DIFF
--- a/morango/models.py
+++ b/morango/models.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 import platform
 import sys
@@ -10,6 +11,10 @@ from django.utils import timezone
 from .certificates import *
 from .manager import SyncableModelManager
 from .utils.uuids import UUIDField, UUIDModelMixin
+
+
+def sha2_uuid(*args):
+    return hashlib.sha256("::".join(args)).hexdigest()[:32]
 
 
 class DatabaseManager(models.Manager):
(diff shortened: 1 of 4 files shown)


morango/models.py
```diff
@@ -238,13 +245,19 @@ class SyncableModel(UUIDModelMixin):
     their data syncable across devices.
     """
 
+    # constant value to insert into partition strings in place of current model's ID, as needed (to avoid circularity)
+    ID_PLACEHOLDER = "<id>"
```
jamalex: Let's actually replace this with `${id}`, so it follows the same `Template` format as `*_scope_def` strings.

morango/models.py
```diff
@@ -238,13 +245,19 @@ class SyncableModel(UUIDModelMixin):
...
     _morango_internal_fields_not_to_serialize = ('_morango_dirty_bit',)
     morango_fields_not_to_serialize = ()
+    morango_profile = None
 
     # morango specific field used for tracking model changes
     _morango_dirty_bit = models.BooleanField(default=True)
     # morango specific field used to store random uuid or unique together fields
     _morango_source_id = models.CharField(max_length=96)
+    # morango specific field used to store the partition on the model
+    _morango_partition = models.TextField()
```
jamalex: `CharField` should be enough for this -- max_length 128, perhaps?

morango/models.py
```diff
@@ -290,18 +303,23 @@ def deserialize(cls, dict_model):
     def merge_conflict(cls, current, incoming):
         return incoming
 
-    def get_partition(self, *args, **kwargs):
-        """Should return a dictionary with any relevant partition keys included, along with their values."""
-        raise NotImplemented("You must define a 'get_partition_names' method on models that inherit from SyncableModel.")
-
+    def calculate_source_id(self):
+        """Should return a string that uniquely defines the model or `None` for a random uuid."""
```
jamalex: "model" -> "model instance"

tests/testapp/facility_profile/models.py
```diff
@@ -32,9 +32,11 @@ class Facility(MorangoMPTTModel, FacilityDataSyncableModel):
     now_date = models.DateTimeField(default=timezone.now)
     parent = TreeForeignKey('self', null=True, blank=True, related_name='children', db_index=True)
 
-    def get_partition(self, *args, **kwargs):
-        return {}
+    def calculate_source_id(self, *args, **kwargs):
+        return ''
```
jamalex: Why empty string?

jamalex: (or is that just temporary until the tests are fixed?)

ralphiee22: temp until tests are fixed.

PR #12, opened 2017-05-10
Add in model name to uuid calc.

ralphiee22:


morango/utils/uuids.py
```diff
@@ -50,6 +50,21 @@ def to_python(self, value):
             return value.hex
         return value
 
+    def get_default(self):
```
ralphiee22: Returned `None` for default because it would return empty string otherwise, and `calculate_uuid` method would not be called.

jamalex: Ooooh.... I wonder... should we just call `calculate_uuid` from in *here*....?

PR #16, opened 2017-05-19
Added generate_root_certificate method and a number of small fixes, along with new and updated tests

jamalex:


morango/certificates.py
```diff
@@ -34,9 +37,56 @@ class Certificate(mptt.models.MPTTModel, UUIDModelMixin):
...
     # when we own a certificate, we'll have the private key for it (otherwise not)
-    private_key = PrivateKeyField(blank=True, null=True)
+    _private_key = PrivateKeyField(blank=True, null=True, db_column="private_key")
+
+    @property
+    def private_key(self):
+        return self._private_key
+
+    @private_key.setter
+    def private_key(self, value):
```
ralphiee22: should we add validation at this level?
or just allow the database to catch it at `save` when we see its an invalid key?

morango/crypto.py
```diff
@@ -168,18 +114,170 @@ def _add_pem_headers(self, pem_string, header_string):
...
-    try:
-        message = message.encode("utf-8", "replace")
-    except UnicodeDecodeError:
-        pass
-    return hashlib.sha256(message).digest()
-    
+class PythonRSAKey(BaseKey):
+
+    _public_key = None
+    _private_key = None
```
ralphiee22: why not put this in the base key class?

jamalex: I did originally, then realized they were only being referenced in the child classes, and were more of an internal implementation detail, as not all "backends" (subclasses) might implement this in the same way (e.g. some might have a single internal key object that stores both the public and private components).

morango/certificates.py
jamalex: I think it should happen at the later validation level -- throwing validation errors when a field's value is set probably isn't expected.

PR #17, opened 2017-06-30
Make platform detection more robust, and add optional system_id input

jamalex:


morango/models.py
```diff
@@ -69,18 +69,26 @@ class InstanceIDModel(UUIDModelMixin):
...
 
     @staticmethod
     def get_or_create_current_instance():
         """Get the instance model corresponding to the current system, or create a new
         one if the system is new or its properties have changed (e.g. OS from upgrade)."""
 
+        # on Android, platform.platform() barfs, so we handle that safely here
+        try:
+            plat = platform.platform()
+        except:
```
ralphiee22: Can we watch a more specific exception here?

jamalex: The exception is pretty generic, and also doesn't make much sense for the context (something like "IOError: this directory is a folder"), so I don't think it would add much semantic value. And if something as simple as `platform.platform()` fails, no matter what the reason, I think we'd want to just handle it gracefully anyway.

PR #19, opened 2017-07-07
Added app_to_store serialization and rmc tests.

ralphiee22:
## Summary

Wrote tests for app to store serialization. Making sure fields are getting updated correctly. Also wrote tests for record max counters updating correctly during serialization. 

## TODO

## Reviewer guidance

If you can think of any more scenarios to test that would be great!


tests/testapp/tests/test_controller_methods.py
```diff
@@ -0,0 +1,254 @@
...
+from morango.models import DatabaseIDModel, DeletedModels, InstanceIDModel, RecordMaxCounter, Store
+
+
+def serialized_facility_factory(identifier):
+    serialized = {'name': "Facility {}".format(identifier),
+                  'id': identifier,
+                  '_morango_source_id': uuid.uuid4().hex,
+                  '_morango_partition': "",
+                  '_morango_dirty_bit': False}
+    return json.dumps(serialized)
```
jamalex: May be easier/more maintainable to create a Facility instance (without saving it), and then calling `serialize` on it?

tests/testapp/tests/test_controller_methods.py
```diff
@@ -0,0 +1,254 @@
...
+        DatabaseIDModel.objects.create()
+        (self.current_id, _) = InstanceIDModel.get_or_create_current_instance()
+        self.range = 10
+        self.mc = MorangoProfileController('facilitydata')
+        for i in range(self.range):
+            self.ident = uuid.uuid4().hex
+            StoreModelFacilityFactory(pk=self.ident, serialized=serialized_facility_factory(self.ident))
+
+    def test_dirty_store_records_are_deserialized(self):
+        self.assertFalse(Facility.objects.all())
```
jamalex: Here and elsewhere, when doing this `assertFalse` type of thing on querysets, might be clearer, and marginally faster, for the queryset to use `.exists()`.

tests/testapp/tests/test_controller_methods.py
```diff
@@ -0,0 +1,254 @@
...
+        Store.objects.update(dirty_bit=False)
+        self.mc._store_to_app()
+        self.assertFalse(Facility.objects.all())
+
+    def test_deleted_models_do_not_get_deserialized(self):
+        Store.objects.update_or_create(defaults={'deleted': True}, id=self.ident)
+        self.mc._store_to_app()
+        with self.assertRaises(Facility.DoesNotExist):
+            Facility.objects.get(id=self.ident)
```
jamalex: Ah, interesting -- we *do* actually want models marked as deleted (and dirty) in the store to then lead to deletion in the app. This test case is compatible with that being the case, but doesn't test it directly.

morango/controller.py
```diff
@@ -69,28 +69,28 @@ def _serialize_into_store(self):
...
-    def open_network_sync_connection(host, scope):
-        pass
-
-    def open_disk_sync_connection(path, scope):
-        pass
+            for store_model in Store.objects.filter(model_name=model_name, dirty_bit=True):
+                if store_model.deleted:
+                    klass_model.objects.filter(id=store_model.id).delete()
+                else:
+                    app_model = klass_model.deserialize(json.loads(store_model.serialized))
```
jamalex: Let's discuss this as part of a question I added on Trello (re: what if app model too is dirty?)

morango/controller.py
```diff
@@ -69,28 +69,28 @@ def _serialize_into_store(self):
...
+                    try:
+                        app_model.save(update_dirty_bit_to=False)
+                    # when deserializing, we catch this exception in case we have a reference to a missing object (foreign key) not in the app layer
+                    except ObjectDoesNotExist:
+                        DeletedModels.objects.update_or_create(defaults={'id': app_model.id, 'profile': app_model.morango_profile},
+                                                               id=app_model.id)
```
jamalex: Still feel like having this logic on BaseSyncableModel could be good, so we don't need to duplicate it. Can discuss.

morango/controller.py
```diff
@@ -69,28 +69,28 @@ def _serialize_into_store(self):
...
+                store_model.dirty_bit = False
+                store_model.save(update_fields=['dirty_bit'])
```
jamalex: As long as transactions are doing everything they're meant to, could we just do a bulk `.update(dirty_bit=False)` at the end?

morango/controller.py
```diff
@@ -7,7 +7,7 @@
 from django.utils.six import iteritems
 from morango.models import DeletedModels, InstanceIDModel, RecordMaxCounter, Store
 
-from .register_models import _profile_models
+from morango.utils.register_models import _profile_models
```
ralphiee22: It is not relative to that location anymore.

morango/controller.py
```diff
@@ -33,8 +33,13 @@ def serialize_into_store(self):
             for (_, klass_model) in iteritems(syncable_dict):
                 for app_model in klass_model.objects.filter(_morango_dirty_bit=True):
                     try:
-                        # set new serialized data on this store model
                         store_model = Store.objects.get(id=app_model.id)
+
+                        # if store record dirty and app record dirty, append store serialized to conflicting data
+                        if store_model.dirty_bit:
+                            store_model.conflicting_serialized_data = store_model.serialized + "\n" + store_model.conflicting_serialized_data
```
jamalex: Is this consistent with the ordering the dequeuing merge conflict appending does? I had imagined it appending the new conflicting records to the end.

morango/controller.py
```diff
@@ -73,24 +78,27 @@ def serialize_into_store(self):
...
-                    try:
-                        app_model.save(update_dirty_bit_to=False)
-                    # when deserializing, we catch this exception in case we have a reference to a missing object (foreign key) not in the app layer
-                    except ObjectDoesNotExist:
-                        DeletedModels.objects.update_or_create(defaults={'id': app_model.id, 'profile': app_model.morango_profile},
-                                                               id=app_model.id)
-                store_model.dirty_bit = False
-                store_model.save(update_fields=['dirty_bit'])
+
+        self.serialize_into_store()
```
jamalex: I think we need this to happen within the `transaction.atomic()`, to ensure nothing gets modified in the app between serialization and deserialization.

morango/models.py
```diff
@@ -271,6 +271,10 @@ class SyncableModel(UUIDModelMixin):
     class Meta:
         abstract = True
 
+    def _update_deleted_models(self):
```
jamalex: Maybe `_mark_as_deleted`? "update" suggests something broader to me

morango/controller.py
```diff
@@ -53,7 +53,7 @@ def serialize_into_store(self):
                         store_model.last_saved_counter = current_id.counter
 
                         # update fields for this store model
-                        store_model.save(update_fields=['serialized', 'last_saved_instance', 'last_saved_counter'])
+                        store_model.save(update_fields=['serialized', 'last_saved_instance', 'last_saved_counter', 'conflicting_serialized_data'])
```
jamalex: What about `deleted`? Or does that get changed elsewhere?

tests/testapp/tests/test_controller_methods.py
```diff
@@ -139,6 +139,32 @@ def test_cascading_delete_updates_store_deleted_flag(self):
...
+        self.mc.serialize_into_store()
+
+        # add serialized fields to conflicting data
+        conflicting.insert(0, serialized)
+        conflicting.insert(0, json.dumps(user.serialize()))
+
+        # set store record and app record dirty bits to true to force serialization merge conflict
+        Store.objects.update(conflicting_serialized_data=serialized, dirty_bit=True)
+        user.username = "user1"
+        user.save(update_dirty_bit_to=True)
```
jamalex: This should be the default, no? (`update_dirty_bit_to=True`)

morango/controller.py
ralphiee22: couple of lines below, where the `DeletedModels` comes into play

tests/testapp/tests/test_controller_methods.py
ralphiee22: this is true...

morango/models.py
ralphiee22: ok

morango/controller.py
jamalex: Ah, cool, see it there now

PR #21, opened 2017-08-02
Add API endpoints for certificates, CSR, syncsession, transfersession, and buffer

jamalex:
## TODO


morango/migrations/0010_auto_20170807_0123.py
```diff
@@ -0,0 +1,120 @@
...
+        ),
+        migrations.AddField(
+            model_name='syncsession',
+            name='remote_ip',
+            field=models.CharField(blank=True, max_length=100),
+        ),
+        migrations.AddField(
+            model_name='transfersession',
+            name='last_activity_timestamp',
+            field=models.DateTimeField(blank=True, default=datetime.datetime(2017, 8, 7, 1, 23, 10, 795426, tzinfo=utc)),
```
ralphiee22: Why is there a default when one was not set? Would it be valid to say that the last activity was when it was created? Is activity only labeled as in the state of pulling or pushing records?

morango/api/viewsets.py
```diff
@@ -0,0 +1,273 @@
...
+        transfersession.save()
+
+    def get_queryset(self):
+        return models.TransferSession.objects.filter(active=True)
+
+
+class BufferViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    permission_classes = (permissions.BufferPermissions,)
+    serializer_class = serializers.BufferSerializer
+    pagination_class = pagination.LimitOffsetPagination
```
ralphiee22: Is the default page size 100 if not set explicitly? Maybe we can have it be explicitly defined somewhere to be a bit clearer.

morango/api/viewsets.py
```diff
@@ -0,0 +1,273 @@
...
+        serial_data = serializers.BufferSerializer(data=data, many=True)
+        if serial_data.is_valid():
+
+            # ensure the transfer session allows pushes, and is same across records
+            if not serial_data.validated_data[0]["transfer_session"].incoming:
+                return response.Response(
+                    "Specified TransferSession does not allow pushes.",
+                    status=status.HTTP_403_FORBIDDEN
+                )
+            if len(set(rec["transfer_session"] for rec in serial_data.data)) > 1:
```
jamalex: In part it's a sanity check, but it's also for security reasons, as in the check on line 251 above we only check against the first record's transfer_session.

morango/certificates.py
```diff
@@ -137,12 +150,53 @@ def check_certificate(self):
...
+        cert = cls.deserialize(cert_data["serialized"], cert_data["signature"])
+
+        # verify the id of the cert matches the id of the outer serialized data
+        assert cert_data["id"] == cert.id
+
+        # check that the expected ID matches, if specified
+        if expected_last_id:
+            assert cert.id == expected_last_id
+
+        # if cert already exists locally, it's already been verified, so no need to continue
```
jamalex: Yep, otherwise the `parent` foreignkeys would have failed. We (can) only save certs for which we have the parent and it itself is already saved.

morango/migrations/0010_auto_20170807_0123.py
```diff
@@ -0,0 +1,120 @@
...
+            name='chunksize',
+        ),
+        migrations.RemoveField(
+            model_name='transfersession',
+            name='records_remaining',
+        ),
+        migrations.AddField(
+            model_name='buffer',
+            name='source_id',
+            field=models.CharField(default=datetime.datetime(2017, 8, 7, 1, 22, 33, 701906, tzinfo=utc), max_length=96),
```
jamalex: MAGIC!

I think it required a default value, and it asked me, and it told me you can use `datetime.datetime()`, so I... did? :D

My plan is to delete all these new migrations and create a fresh slate before we integrate into Kolibri.

morango/migrations/0010_auto_20170807_0123.py
jamalex: Since there are not yet any transfer sessions "in the wild", this will never actually get used. It asked me for a default for purposes of migration, and I gave it one.

morango/models.py
```diff
@@ -113,20 +122,33 @@ class SyncSession(models.Model):
...
-    connection_path = models.CharField(max_length=1000)  # file path if kind=disk, and base URL if kind=network
-    connection_params = models.TextField(default=u"{}")
+    connection_path = models.CharField(max_length=1000)  # file path if kind=disk, and base URL of server if kind=network
+
+    # for network connections, keep track of the IPs on either end
+    local_ip = models.CharField(max_length=100, blank=True)
+    remote_ip = models.CharField(max_length=100, blank=True)
+
+    local_instance = models.TextField(default=u"{}")
+    remote_instance = models.TextField(default=u"{}")
```
jamalex: This is the full serialized instance (all the input params), for tracking/debugging purposes.

The instance won't change without a server restart, which in general shouldn't happen in the midst of a sync, but in principle could (especially on the server side). But since the only reason we track it here is to support debugging/tracking, it wouldn't cause problems if it changed.

morango/models.py
```diff
@@ -155,24 +155,18 @@ class TransferSession(models.Model):
     """
 
     id = UUIDField(primary_key=True)
-    # partition/filter to know what subset of data is to be synced
-    filter = models.TextField()
-    # is session pushing or pulling data
-    incoming = models.BooleanField()
-    # is this session actively pushing or pulling data?
-    active = models.BooleanField(default=True)
-    chunksize = models.IntegerField(default=500)
```
jamalex: For now we'll use a default. Then we can make that configurable as a Django setting. Then we can have a "pullback" scheme for reducing it in the case of network timeouts etc.

I didn't see any reason to store it, since it's all managed by the client side (whether pushing or pulling), and it maintains its own state in memory.

morango/models.py
```diff
@@ -191,16 +219,19 @@ class Store(AbstractStore):
...
+class Buffer(AbstractStore, UUIDModelMixin):
     """
-    ``Buffer`` is where records from the internal store are kept temporarily,
-    until they are sent or received by another morango instance.
+    ``Buffer`` is where records from the internal store are queued up temporarily, before being
+    sent to another morango instance, or stored while being received from another instance, before
+    dequeuing into the local store.
     """
 
+    uuid_input_fields = ("transfer_session_id", "model_uuid")
```
jamalex: Main reason was to avoid accidental duplication of buffer records if the same chunk was pushed/pulled twice.

Probably could achieve the same objective with `unique_together`? Any other ideas?

tests/testapp/tests/test_api.py
```diff
@@ -0,0 +1,791 @@
...
+        self.test_syncsession_can_be_created()
+
+        syncsession = SyncSession.objects.get()
+        self.assertEqual(syncsession.active, True)
+
+        response = self.client.delete(reverse('syncsessions-detail', kwargs={"pk": syncsession.id}), format='json')
+        self.assertEqual(response.status_code, 204)
+
+        # check that the syncsession was "deleted"
+        self.assertEqual(SyncSession.objects.get().active, False)
```
jamalex: If 0 records exist, it blows up. If > 1 records exist, it blows up. Otherwise, it returns the single record that does exist.

Will change though, yep, if it makes it clearer.

tests/testapp/tests/test_api.py
```diff
@@ -584,3 +586,206 @@ def test_inactive_transfersession_cannot_be_deleted(self):
...
+            # check that the buffer items were not created
+            self.assertEqual(Buffer.objects.count(), 0)
+
+    def test_push_valid_buffer_chunk(self):
+        rec_1 = self.build_buffer_item(push=True, filter=self.default_push_filter)
+        rec_2 = self.build_buffer_item(transfer_session=rec_1.transfer_session)
+        rec_3 = self.build_buffer_item(transfer_session=rec_1.transfer_session)
+        self.make_buffer_post_request([rec_1, rec_2, rec_3], expected_status=201)
+
+    def test_push_with_invalid_model_uuid(self):
```
jamalex: Do it right or don't do it at all! :1st_place_medal: 

Why go on? What's left to live for, if a record has failed ye?

jamalex: (Given we wouldn't be able to complete the session anyway, due to snapshotting concerns)

PR #22, opened 2017-08-16
Queuing and Dequeuing and tests

ralphiee22:
## Summary

Logic for queuing up records from the store into the buffer. Logic for dequeuing records from the store into the buffer. Logic for calculating FilterMaxCounters. Tests have been written which cover all these operations. 

## TODO

## Reviewer guidance


morango/syncsession.py
```diff
@@ -0,0 +1,407 @@
...
+        with connection.cursor() as cursor:
+            self._dequeuing_delete_rmcb_records(cursor)
+            self._dequeuing_delete_buffered_records(cursor)
+            self._dequeuing_merge_conflict_rmcb(cursor)
+            current_id = InstanceIDModel.get_and_update_current_instance()
+            self._dequeuing_merge_conflict_buffer(cursor, current_id)
+            self._dequeuing_update_rmcs_last_saved_by(cursor, current_id)
+            self._dequeuing_delete_mc_rmcb(cursor)
+            self._dequeuing_delete_mc_buffer(cursor)
+            self._dequeuing_insert_remaining_buffer(cursor)
```
ralphiee22: After deleting the merge conflicting buffers, all we have is fast forwards left and records were never in the store before. So I forego doing the fast forward queries and just insert the remaining records into the store.

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+        buffer_ids = Buffer.objects.values_list('model_uuid', flat=True)
+        rmcb_ids = RecordMaxCounterBuffer.objects.values_list('model_uuid', flat=True)
+        self.assertNotIn(self.data['user2'].id, buffer_ids)
+        self.assertNotIn(self.data['user2'].id, rmcb_ids)
+        for i in self.data['user3_sumlogs']:
+            self.assertIn(i.id, buffer_ids)
+            self.assertIn(i.id, rmcb_ids)
+        for i in self.data['user3_interlogs']:
+            self.assertIn(i.id, buffer_ids)
+            self.assertIn(i.id, rmcb_ids)
```
jamalex: Perhaps the reusable function I described above could be more general, along the lines of:

`assertRecordsBuffered(filter, fsics, records)`

which here would be:
```
filter_prefixes = ['{}:user:summary'.format(self.data['user3'].id), '{}:user:interaction'.format(self.data['user3'].id)]
self.assertRecordsBuffered(filter_prefixes, self.fsics, self.data['user3_sumlogs'] + self.data['user3_interlogs'])
```

That could cut down the verbosity and increase the expressiveness of these tests.

jamalex: And if these are exhaustive lists of what should be buffered, could even just do something like `self.assertSetEqual`, and then could also use it for `test_fsic_counters_too_high` (with an empty list).

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+    def setUp(self):
+        self.data = {}
+        DatabaseIDModel.objects.create()
+        (self.current_id, _) = InstanceIDModel.get_or_create_current_instance()
+
+        # create controllers for app/store/buffer operations
+        self.data['mc'] = MorangoProfileController('facilitydata')
+        self.data['sc'] = SyncClient('host', 'facilitydata')
+
+        self.data.update(create_buffer_and_store_dummy_data(self.data['sc'].transfer_session_id))
```
jamalex: Isn't `transfer_session_id` still `None` at this point?

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def test_dequeuing_merge_conflict_rmc(self):
+        rmc = RecordMaxCounter.objects.get(instance_id=self.data['model2_rmc_ids'][0], store_model_id=self.data['model2'])
+        rmcb = RecordMaxCounterBuffer.objects.get(instance_id=self.data['model2_rmc_ids'][0], model_uuid=self.data['model2'])
+        self.assertNotEqual(rmc.counter, rmcb.counter)
+        with connection.cursor() as cursor:
+            self.data['sc']._dequeuing_merge_conflict_rmc(cursor)
+        rmc = RecordMaxCounter.objects.get(instance_id=self.data['model2_rmc_ids'][0], store_model_id=self.data['model2'])
+        rmcb = RecordMaxCounterBuffer.objects.get(instance_id=self.data['model2_rmc_ids'][0], model_uuid=self.data['model2'])
+        self.assertEqual(rmc.counter, rmcb.counter)
```
jamalex: Seems like there are three cases that would be good to test; not certain whether all are captured here:
- rmcb.instance_id isn't in rmc
- rmcb.counter is less than or equal to rmc.counter
- rmcb.counter is greater than rmc.counter

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+    def test_dequeuing_ff_rmc(self):
+        # this only transfers rmcs that have the same instance id (does not handle instance ids it has not seen before)
+        rmc = RecordMaxCounter.objects.get(instance_id=self.data['model3_rmc_ids'][0], store_model_id=self.data['model3'])
+        rmcb = RecordMaxCounterBuffer.objects.get(instance_id=self.data['model3_rmcb_ids'][0], model_uuid=self.data['model3'])
+        self.assertNotEqual(rmc.counter, rmcb.counter)
+        with connection.cursor() as cursor:
+            self.data['sc']._dequeuing_ff_rmc(cursor)
+        rmc = RecordMaxCounter.objects.get(instance_id=self.data['model3_rmc_ids'][0], store_model_id=self.data['model3'])
+        rmcb = RecordMaxCounterBuffer.objects.get(instance_id=self.data['model3_rmcb_ids'][0], model_uuid=self.data['model3'])
+        self.assertEqual(rmc.counter, rmcb.counter)
```
jamalex: Looks like this tests the "update" case, there's also the case of the instance ID not yet existing in the RMC ("insert").

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def test_dequeuing_ff_store(self):
+        store_model = Store.objects.get(id=self.data['model3'])
+        buffer_model = Buffer.objects.get(model_uuid=self.data['model3'])
+        self.assertNotEqual(store_model.last_saved_instance, buffer_model.last_saved_instance)
+        with connection.cursor() as cursor:
+            self.data['sc']._dequeuing_ff_store(cursor)
+        store_model = Store.objects.get(id=self.data['model3'])
+        buffer_model = Buffer.objects.get(model_uuid=self.data['model3'])
+        self.assertEqual(store_model.last_saved_instance, buffer_model.last_saved_instance)
```
jamalex: Perhaps also assert on a couple of other key fields, e.g. the serialized data, to make sure it made it across

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def test_dequeuing_delete_ff_buffer(self):
+        self.assertTrue(Buffer.objects.filter(model_uuid=self.data['model3']).exists())
+        with connection.cursor() as cursor:
+            self.data['sc']._dequeuing_delete_ff_buffer(cursor)
+        self.assertFalse(Buffer.objects.filter(model_uuid=self.data['model3']).exists())
```
jamalex: Perhaps asserting that certain things *weren't* deleted? (e.g. any new records)

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def test_dequeuing_delete_ff_rmcb(self):
+        self.assertTrue(RecordMaxCounterBuffer.objects.filter(instance_id=self.data['model3_rmcb_ids'][0], model_uuid=self.data['model3']).exists())
+        with connection.cursor() as cursor:
+            self.data['sc']._dequeuing_delete_ff_rmcb(cursor)
+        self.assertFalse(RecordMaxCounterBuffer.objects.filter(instance_id=self.data['model3_rmcb_ids'][0], model_uuid=self.data['model3']).exists())
```
jamalex: Same here?

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def test_dequeuing_insert_remaining_buffer(self):
+        self.assertFalse(Store.objects.filter(id=self.data['model4']).exists())
+        with connection.cursor() as cursor:
+            self.data['sc']._dequeuing_insert_remaining_buffer(cursor)
+        self.assertTrue(Store.objects.filter(id=self.data['model4']).exists())
+
+    def test_dequeuing_insert_remaining_rmcb(self):
+        for i in self.data['model4_rmcb_ids']:
+            self.assertFalse(RecordMaxCounter.objects.filter(instance_id=i, store_model_id=self.data['model4']).exists())
```
jamalex: Could you get rid of the loop and just do `instance_id__in=self.data['model4_rmcb_ids']`?

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+        with connection.cursor() as cursor:
+            self.data['sc']._dequeuing_insert_remaining_rmcb(cursor)
+        for i in self.data['model4_rmcb_ids']:
+            self.assertTrue(RecordMaxCounter.objects.filter(instance_id=i, store_model_id=self.data['model4']).exists())
```
jamalex: In this case, if collapsing as suggested above, you'd want to assert the `count()` value

jamalex:
Another general comment:

Would be nice to write a test for the full `_integrate_into_store` method, explicitly checking all the conditions necessary for the full desired end state of the Buffer, RMCBuffer, Store, and RMC.

morango/models.py
```diff
@@ -222,11 +229,40 @@ class Meta:
...
-    of those filters, as well as the maximum counter we received for each instance during the sync session.
+    instances for a particular partition prefix. Whenever 2 morango instances sync with each other we keep track
+    of those partition prefixes from the filters, as well as the maximum counter we received for each instance during the sync session.
     """
 
-    filter = models.TextField()
+    partition = models.CharField(max_length=128, default="")
+
+    @classmethod
+    def calculate_filter_max_counters(cls, filters):
```
jamalex: After merging https://github.com/learningequality/morango/pull/21, this can be updated to take an instance of `Filter`, and then we won't need to split on "\n".

morango/syncsession.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+class SyncClient(object):
+    """
+    Controller to support client in initiating syncing and performing related operations.
+    """
+    def __init__(self, host, profile):
+        self.host = host
+        self.profile = profile
+        self.transfer_session_id = uuid.uuid4().hex
```
ralphiee22: ahh yes, mainly had it for testing purposes.

morango/syncsession.py
```diff
@@ -0,0 +1,407 @@
...
+
+            # transfer buffer serialized into conflicting store
+            merge_conflict_store = """REPLACE INTO {store} (id, serialized, deleted, last_saved_instance, last_saved_counter, model_name,
+                                                            profile, partition, conflicting_serialized_data, dirty_bit)
+                                                SELECT store.id, store.serialized, store.deleted, '{current_instance_id}', {current_instance_counter},
+                                                       store.model_name, store.profile, store.partition,
+                                                       buffer.serialized || '\n' || store.conflicting_serialized_data, 1
+                                                FROM {buffer} AS buffer, {rmcb} AS rmcb, {store} AS store, {rmc} as rmc
+                                    /*Scope to a single record.*/
+                                    WHERE store.id = rmcb.model_uuid
```
ralphiee22: It is still checking the conditions for a merge conflict which depend on the rmcb.

morango/syncsession.py
```diff
@@ -0,0 +1,407 @@
...
+                                          """.format(buffer=Buffer._meta.db_table,
+                                                     rmcb=RecordMaxCounterBuffer._meta.db_table,
+                                                     store=Store._meta.db_table,
+                                                     rmc=RecordMaxCounter._meta.db_table,
+                                                     transfer_session_id=self.transfer_session_id,
+                                                     current_instance_id=current_id.id,
+                                                     current_instance_counter=current_id.counter)
+            cursor.execute(merge_conflict_store)
+
+    def _dequeuing_update_rmcs_last_saved_by(self, cursor, current_id):
```
ralphiee22: I pass it around as not to create a new connection each time I want to execute a query. Below I call each function separately and wrap the whole thing in a transaction.

morango/syncsession.py
```diff
@@ -0,0 +1,407 @@
...
+
+            # updated or create rmc for merge conflicts with id of current system
+            merge_conflict_store = """REPLACE INTO {rmc} (instance_id, counter, store_model_id)
```
ralphiee22: Yes, if it violates the `unique_together` constraint it will replace that row with the new values. In this case, the new counter.

morango/syncsession.py
```diff
@@ -0,0 +1,407 @@
...
+                                                                                  AND store.last_saved_counter <= rmcb2.counter
+                                                                                  AND rmcb2.transfer_session_id = '{transfer_session_id}'))
+                               """.format(buffer=Buffer._meta.db_table,
+                                          store=Store._meta.db_table,
+                                          rmc=RecordMaxCounter._meta.db_table,
+                                          rmcb=RecordMaxCounterBuffer._meta.db_table,
+                                          transfer_session_id=self.transfer_session_id)
+        cursor.execute(delete_mc_rmc)
+
+    def _dequeuing_ff_rmc(self, cursor):
```
ralphiee22: Yes, I should be able to delete this method and the `_dequeuing_ff_store`.

tests/testapp/tests/test_models.py
```diff
@@ -0,0 +1,57 @@
...
+        DatabaseMaxCounterFactory(instance_id=self.instance_b, partition=self.prefix_b, counter=12)
+        DatabaseMaxCounterFactory(instance_id=self.instance_b, partition=self.user_prefix_b, counter=5)
+        DatabaseMaxCounterFactory(instance_id=self.instance_b, partition=self.user2_prefix_b, counter=2)
+
+    def test_filter_not_in_dmc(self):
+        fmcs = DatabaseMaxCounter.calculate_filter_max_counters("ZZZ")
+        self.assertFalse(fmcs)
+
+    def test_instances_for_one_partition_but_not_other(self):
+        fmcs = dict(DatabaseMaxCounter.calculate_filter_max_counters(self.user_prefix_a + "\n" + self.user_prefix_b))
```
ralphiee22: its returned as a list, so maybe just always coerce to dict but in the actual function call.

tests/testapp/tests/test_syncsession_methods.py
```diff
@@ -0,0 +1,251 @@
...
+        rmcb_ids = RecordMaxCounterBuffer.objects.values_list('model_uuid', flat=True)
+        for i in self.data['group1_c1']:
+            self.assertIn(i.id, buffer_ids)
+            self.assertIn(i.id, rmcb_ids)
+        for i in self.data['group1_c2']:
+            self.assertIn(i.id, buffer_ids)
+            self.assertIn(i.id, rmcb_ids)
+        for i in self.data['group2_c1']:
+            self.assertIn(i.id, buffer_ids)
+            self.assertIn(i.id, rmcb_ids)
```
ralphiee22: My reason for doing several loops was because in case one of the data groups is not actually in the list of ids, then I actually know which group is failing. Since each group has different partition values and last_saved_bys. Instead of grouping them together and having it all fail if only one of the groups is wrong.

tests/testapp/tests/test_syncsession_methods.py
ralphiee22: Well since I set it at instantiation of the `SyncClient`, but will need to update tests once we change that.

tests/testapp/tests/test_syncsession_methods.py
ralphiee22: Deleted this test since we are not using it, but since we just push all remaining records into the store (after all the deletions), then those instance IDs that did not exist before are now in the RMC.

tests/testapp/tests/test_syncsession_methods.py
ralphiee22: Also in regards to the comment above, I don't like using `count()` because it could be the case that other records not meant to be in the result set are actually there. So we could have count be 3, but the records are not the 3 we want. So hence the loop is to at least be sure that the records we want to be in there are actually in there. There should be other assertions that assert certain data isn't in there. @jamalex

tests/testapp/tests/test_syncsession_methods.py
ralphiee22: I think the first case is handled after we just merge the remaining records into the store and RMC.

tests/testapp/tests/test_syncsession_methods.py
jamalex: Ah, fair point. Just thinking of ways to make it less verbose. Could perhaps be a helper method like `self.assertAllIDsPresent(self.data['group2_c1'], buffer_ids, rmcb_ids)` or something

PR #23, opened 2017-08-31
Stop using proxied_children in django internal api.

ralphiee22:
## Summary

`Proxied_children` is an internal django field that was used but later removed. We still use the internal api for registering proxy models, but the methods are consistent across latest versions of django.


ralphiee22:
We probably won't have to do special logic for ordering the proxy models that have self referencing FKs because of https://github.com/learningequality/morango/pull/25. When deserializing we will start at the root of trees and work our way down.

jamalex:
Nice!

PR #24, opened 2017-09-06
Filtered serialization

ralphiee22:
## Summary

If a `filter_id` is passed in, we only serialize records whose partition values contain that id value.
Also, did a bulk_create operation for creating new store and rmc records.


morango/controller.py
```diff
@@ -15,7 +17,7 @@ def __init__(self, profile):
         assert profile, "profile needs to be defined."
         self.profile = profile
 
-    def serialize_into_store(self):
+    def serialize_into_store(self, filters=None):
```
jamalex: Perhaps nice to have this argument be an actual `Filter` class instance.

jamalex: If you don't think that makes sense, then better to rename arg to `filter_prefixes`

morango/controller.py
```diff
@@ -25,10 +27,20 @@ def serialize_into_store(self):
...
+            # create Q objects for filtering by prefixes
+            if filters:
+                prefix_condition = functools.reduce(lambda x, y: x | y, [Q(_morango_partition__startswith=prefix) for prefix in filters])
+
             # filter through all models with the dirty bit turned on
             syncable_dict = _profile_models[self.profile]
             for (_, klass_model) in iteritems(syncable_dict):
-                for app_model in klass_model.objects.filter(_morango_dirty_bit=True):
+                klass_queryset = klass_model.objects.filter(_morango_dirty_bit=True)
+                if filters:
```
jamalex: Might be more semantic to conditionalize on `prefix_condition` here (might mean setting it to `None` by default at the top)

tests/testapp/tests/test_controller_methods.py
```diff
@@ -164,6 +165,23 @@ def test_conflicting_data_appended(self):
...
+        self.assertFalse(Store.objects.filter(id=fac.id).exists())
+        self.assertTrue(Store.objects.filter(id=user.id).exists())
+        self.assertTrue(Store.objects.filter(id=log.id).exists())
+
+    def test_filtered_serialization_multiple_filter(self):
+        user = MyUser.objects.create(username='deadbeef')
+        user2 = MyUser.objects.create(username='alivebeef')
+        log = SummaryLog.objects.create(user=user)
+        self.mc.serialize_into_store(filters=Filter(user._morango_partition + "\n" + user2._morango_partition))
+        self.assertTrue(Store.objects.filter(id=user2.id).exists())
```
jamalex: Maybe also assert that `fac.id` doesn't exist here, as above

morango/controller.py
jamalex: Ah, nevermind, it *is* a `Filter`, but filters are iterable. I think just `filter` is ideal as the param name here, then.

PR #25, opened 2017-09-13
Safely handle deserializing self referencing FK models

ralphiee22:
## Summary

In the future, we want to handle deserializing models which have self referential foreign keys. This PR adds a `_self_ref_fk` to `store` and `buffer`, which allows us to track where the FK dependencies point to, so we can safely **deserialize** the models in the correct dependency order.

Added a test, but more will be added.
## TODO

- [x] Ensure syncable models only have 1 self referential FK. (when registering models)


morango/controller.py
```diff
@@ -1,14 +1,24 @@
...
 
 
+def _self_referential_fk(klass_model):
+    """
+    Return whether this model has a self ref FK, and the name for the field
+    """
+    for f in klass_model._meta.concrete_fields:
+        if f.related_model == klass_model:
+            return True, f.attname
+    return False, None
```
jamalex: This works, but might be enough to just return the attname or None, since they're appropriately booleanizable themselves?

morango/controller.py
```diff
@@ -62,6 +72,10 @@ def serialize_into_store(self):
                             'profile': app_model.morango_profile,
                             'partition': app_model._morango_partition,
                         }
+                        (self_ref_fk_bool, self_ref_fk_name) = _self_referential_fk(klass_model)
```
jamalex: Maybe a comment above here.

morango/controller.py
```diff
@@ -62,6 +72,10 @@ def serialize_into_store(self):
...
+                        if self_ref_fk_bool:
+                            self_ref_fk_value = getattr(app_model, self_ref_fk_name)
+                            kwargs.update({'_self_ref_fk': self_ref_fk_value if self_ref_fk_value else ''})
```
jamalex: `self_ref_fk_value if self_ref_fk_value else ''` could just be `self_ref_fk_value or ''`

morango/controller.py
jamalex: Also, perhaps good to throw an error if there are two self-references?

tests/testapp/tests/test_controller_methods.py
```diff
@@ -298,3 +298,28 @@ def test_record_with_dirty_bit_off_doesnt_deserialize(self):
...
+        Store.objects.update(dirty_bit=True)
+        grandchild1 = FacilityModelFactory(parent=child2)
+        grandchild2 = FacilityModelFactory(parent=child2)
+
+        self.mc.deserialize_from_store()
+        # ensure tree structure in app layer is correct
+        self.assertTrue(Facility.objects.filter(id=child1.id).exists())
+        self.assertFalse(Facility.objects.filter(id=child2.id).exists())
+        self.assertFalse(Facility.objects.filter(id=grandchild1.id).exists())
+        self.assertFalse(Facility.objects.filter(id=grandchild2.id).exists())
```
jamalex: Any tests needed for the `_self_referential_fk` stuff?

tests/testapp/tests/test_controller_methods.py
```diff
@@ -358,3 +354,18 @@ def test_delete_model_in_store_deletes_models_in_app(self):
...
         self.assertFalse(Facility.objects.filter(id=grandchild1.id).exists())
         self.assertFalse(Facility.objects.filter(id=grandchild2.id).exists())
+
+    def test_models_created_successfully(self):
+        root = FacilityModelFactory()
+        child1 = FacilityModelFactory(parent=root)
+        child2 = FacilityModelFactory(parent=root)
+        self.mc.serialize_into_store()
+        Facility.objects.all().delete()
+        DeletedModels.objects.all().delete()
```
jamalex: Maybe do another serialization in here, otherwise there might be stuff left around in `DeletedModels`?

morango/utils/register_models.py
```diff
@@ -18,6 +18,15 @@
 def _get_foreign_key_classes(m):
     return set([field.rel.to for field in m._meta.fields if isinstance(field, ForeignKey)])
 
+def _multiple_self_ref_fk_check(class_model):
+    self_fk = []
```
jamalex: Simple docstring would be great

morango/utils/register_models.py
```diff
@@ -57,6 +66,8 @@ def add_syncable_models():
         # several validation checks to assert models will be syncing correctly
         if issubclass(model_class, SyncableModel):
             name = model_class.__name__
+            if _multiple_self_ref_fk_check(model_class):
+                raise InvalidMorangoModelConfiguration("Syncing models with more than 1 self referential ForeignKey are not supported.")
```
jamalex: "syncing ... *is* not supported"

morango/models.py
```diff
@@ -235,14 +235,20 @@ class Store(AbstractStore):
...
             klass_model.objects.filter(id=self.id).delete()
         # inflate model and attempt to save
         else:
             app_model = klass_model.deserialize(json.loads(self.serialized))
+            if self_ref_fk:
+                # if the field has a value, we put it on the actual self referencing fk field
+                if self._self_ref_fk:
+                    setattr(app_model, self_ref_fk, self._self_ref_fk)
+                else:
+                    setattr(app_model, self_ref_fk, None)
```
jamalex: Is this not already handled by the deserialization itself? We still include the field in the serialized data, no?

ralphiee22: Not in the case of mptt I believe. We leave out `parent_id`.

PR #27, opened 2017-09-23
Network operations for communication between morango instances

ralphiee22:
## Summary

Adds the necessary network calls to create and retrieve sync sessions, transfer sessions, & certificates. 
Also added unit tests for the different operations.

## TODO

- [ ] Add in logic and tests for pulling and pushing chunks?

## Reviewer guidance


morango/syncsession.py
```diff
@@ -21,21 +33,157 @@ class Connection(object):
...
+        # request certs for this primary partition, where the server also has a private key for
+        remote_certs_resp = self._request(api_urls.CERTIFICATE, params={'primary_partition': primary_partition})
+
+        # inflate remote certs into a list of unsaved models
+        for cert in json.loads(remote_certs_resp.data):
+            remote_certs.append(Certificate.deserialize(cert["serialized"], cert["signature"]))
+        return remote_certs
+
+    def certificate_signing_request(self, parent_cert, scope_definition_id, scope_params, username=None, password=None):
+        csr_key = Key()
```
jamalex: Looks like this key may not actually get saved? (which means we'd end up with a cert we couldn't use)

morango/syncsession.py
```diff
@@ -21,21 +33,157 @@ class Connection(object):
...
+        # build up data for csr
+        data = {
+            "parent": parent_cert.id,
+            "profile": self.profile,
+            "scope_definition": scope_definition_id,
+            "scope_version": parent_cert.scope_version,
+            "scope_params": scope_params,
+            "public_key": csr_key.get_public_key_string()
+        }
+        csr_resp = self._request(api_urls.CERTIFICATE, method="POST", data=data, auth=(username, password))
```
jamalex: The Kolibri basic auth stuff requires that you specify a facility, also -- I guess we can leave it up to the calling code to construct the multi-param username, as that's not general-purpose.

jamalex: (can discuss -- may be useful to handle some of it here, with some helpers, taking in `auth` as a `dict`)

morango/syncsession.py
```diff
@@ -21,21 +33,157 @@ class Connection(object):
...
+            'id': uuid.uuid4().hex,
+            'filter': filter_prefixes,
+            'push': push,
+            'sync_session_id': self.sync_session.id,
+            'local_fsic': DatabaseMaxCounter.calculate_filter_max_counters(filter_prefixes)
+        }
+        transfer_resp = self._request(api_urls.TRANSFERSESSION, method="POST", data=data)
+
+        # if transfersession successfully created, create on our own side
+        if transfer_resp.status_code == status.HTTP_201_CREATED:
```
jamalex: Probably want to raise something if it fails, also.

jamalex: And I think we decided it would be useful to save the transfersession locally before doing the post, since for a PULL it might take quite a while to respond (as it queues things up on the server side), and it's nice to have the transfersession object existing locally to show status while that happens.

morango/syncsession.py
```diff
@@ -21,21 +33,157 @@ class Connection(object):
...
+            data['remote_fsic'] = transfer_resp.data.get('local_fsic')
+            data['start_timestamp'] = timezone.now()
+            data['last_activity_timestamp'] = timezone.now()
+            self.current_transfer_session = TransferSession.objects.create(**data)
```
jamalex: Building on the suggestion to move storage of the syncsession out to the syncclient, let's chat through where it makes sense to have this live.

tests/testapp/tests/test_connection.py
```diff
@@ -0,0 +1,134 @@
...
+            self.network_connection.create_sync_session(self.subset_cert, self.root_cert)
+
+    @mock_patch_decorator
+    def test_get_remote_certs(self):
+        # mock certs being returned by server
+        certs = self.subset_cert.get_ancestors(include_self=True)
+        cert_serialized = json.dumps(CertificateSerializer(certs, many=True).data)
+        Connection._request.return_value = mock.Mock(data=cert_serialized)
+
+        # we want to see if the models are created (not saved) successfully
```
jamalex: Wait, don't they already exist, since you got these from the DB, and didn't delete the ones on disk before saving the new ones?

tests/testapp/tests/test_connection.py
```diff
@@ -0,0 +1,134 @@
...
+    def test_get_cert_chain(self):
+        # mock a cert chain being returned by server
+        certs = self.subset_cert.get_ancestors(include_self=True)
+        cert_serialized = json.dumps(CertificateSerializer(certs, many=True).data)
+        Connection._request.return_value = mock.Mock(data=cert_serialized)
+        Certificate.objects.all().delete()
+
+        # we only want to make sure the cert chain is saved
+        self.network_connection._get_certificate_chain(certs[1])
+        self.assertEqual(Certificate.objects.count(), certs.count())
```
jamalex: This may be trivially true; would be best to evaluate `certs.count()` into a var at the top, rather than after stuff has been deleted.

morango/syncsession.py
```diff
@@ -21,21 +33,157 @@ class Connection(object):
...
-    def _push_chunk(self, data):
-        raise NotImplementedError
+    def _get_certificate_chain(self, server_cert):
+        # get ancestors certificate chain for this server cert
+        cert_chain_resp = self._request(api_urls.CERTIFICATE, params={'ancestors_of': server_cert.id})
+
+        # upon receiving cert chain from server, we attempt to save the chain into our records
+        Certificate.save_certificate_chain(cert_chain_resp.data, expected_last_id=server_cert.id)
+
+    def _create_transfer_session(self, push, filter_prefixes):
```
jamalex: Any reason not to have this be a `Filter` instance?

tests/testapp/tests/test_connection.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @mock_patch_decorator
+    def test_close_transfer_session(self):
+        session = SyncSession.objects.create(id=uuid.uuid4().hex, profile="profile", last_activity_timestamp=timezone.now())
+        self.network_connection.sync_session = session
+        transfer = TransferSession.objects.create(id=uuid.uuid4().hex, filter="", push=True, sync_session=session, last_activity_timestamp=timezone.now())
+        self.network_connection.current_transfer_session = transfer
+
+        self.assertEqual(TransferSession.objects.filter(active=True).count(), 1)
+        self.network_connection._close_transfer_session()
```
jamalex: This is tricky, because it basically tries to delete it twice, once through the ORM and once through the API, right? So we're not fully testing both paths. We could perhaps assert that certain mocked methods are called once, or perhaps we leave fuller testing for "multi-server" mode tests.

morango/syncsession.py
```diff
@@ -21,21 +33,157 @@ class Connection(object):
...
+        self.base_url = base_url
         self.profile = profile
-        self.session_id = uuid.uuid4()
+        self.sync_session = None
         self.current_transfer_session = None
-        # self.connection =
 
-    def get_remote_certificates():
-        pass
+    def _request(self, endpoint, method="GET", lookup=None, data={}, params={}, auth=None):
```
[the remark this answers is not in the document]
ralphiee22: Lookup is basically passing in the `pk` value so we know what model to patch or delete.

morango/syncsession.py
```diff
@@ -21,21 +33,157 @@ class Connection(object):
...
+    def create_sync_session(self, client_cert, server_cert):
+        # if server cert does not exist locally, retrieve it from server
+        if not Certificate.objects.filter(id=server_cert.id).exists():
+            self._get_certificate_chain(server_cert)
+
+        # request the server for a one-time-use nonce
+        nonce_resp = self._request(api_urls.NONCE, method="POST")
+
+        if nonce_resp.status_code == status.HTTP_201_CREATED:
+            nonce = json.loads(nonce_resp.content.decode())["id"]
```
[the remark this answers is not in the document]
ralphiee22: I can pass in an `else`, but I was hoping the `raise_for_status` on the requests would raise any flags.

tests/testapp/tests/test_connection.py
ralphiee22: will just assert the mock method is called.

tests/testapp/tests/test_connection.py
ralphiee22: Done!

tests/testapp/tests/test_connection.py
ralphiee22: I'm taking the "data" from the request and seeing if the cert models are inflated correctly (not saving them) and if they match the original certs "sent".

morango/syncsession.py
ralphiee22: JK!

morango/migrations/0001_initial.py
```diff
@@ -191,7 +191,7 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', morango.utils.uuids.UUIDField(primary_key=True, serialize=False)),
                 ('filter', models.TextField()),
-                ('incoming', models.BooleanField()),
+                ('push', models.BooleanField()),
```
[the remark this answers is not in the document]
ralphiee22: going to replace the migrations that I deleted so this should be resolved then.

morango/syncsession.py
```diff
@@ -134,35 +152,78 @@ def certificate_signing_request(self, parent_cert, scope_definition_id, scope_pa
...
-            'local_fsic': DatabaseMaxCounter.calculate_filter_max_counters(filter_prefixes)
+            'local_fsic': DatabaseMaxCounter.calculate_filter_max_counters(filter)
         }
-        transfer_resp = self._request(api_urls.TRANSFERSESSION, method="POST", data=data)
+        data['start_timestamp'] = timezone.now()
+        data['last_activity_timestamp'] = timezone.now()
+        self.current_transfer_session = TransferSession.objects.create(**data)
+
+        # create transfer session on server side
+        transfer_resp = self.sync_connection._create_transfer_session(self.sync_session)
```
ralphiee22: meant to pass in `data` here

morango/syncsession.py
```diff
@@ -134,35 +152,78 @@ def certificate_signing_request(self, parent_cert, scope_definition_id, scope_pa
...
+    def close_sync_session(self):
+
+        # "delete" sync session on server side
+        self.sync_connection._close_sync_session(self.sync_session)
+
+        # "delete" our own local sync session
+        self.sync_connection.sync_session.active = False
+        self.sync_connection.sync_session.save()
+        self.sync_connection.sync_session = None
+        self = None
```
jamalex: As it's just a local variable, I don't think this does anything?

morango/syncsession.py
jamalex: Was this resolved?

ralphiee22: yes in the next PR.

morango/syncsession.py
ralphiee22: can update.

PR #29, opened 2017-10-18
Migrations fix

ralphiee22:
## Issues addressed

Bring back the previously deleted migrations.
Also added fix for mptt per https://django-mptt.github.io/django-mptt/upgrade.html#id1.


morango/migrations/0001_initial.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Generated by Django 1.9.13 on 2017-09-05 23:22
+# Generated by Django 1.9 on 2017-05-08 23:42
```
jamalex: Apparently near-midnight is witching hour for initial migration creation.

morango/models.py
```diff
@@ -293,6 +295,10 @@ class DatabaseMaxCounter(AbstractCounter):
     """
 
     partition = models.CharField(max_length=128, default="")
+    profile = models.CharField(max_length=40)
```
ralphiee22: Since we serialize based on profile, I think this field is important for tracking that. 
The empty `""` for profile could signify full serialization?
we can also just run through all the profiles and have the partition be the default value `""` to mean we serialized everything for each profile.

morango/models.py
```diff
@@ -293,6 +295,10 @@ class DatabaseMaxCounter(AbstractCounter):
...
+
+    class Meta:
+        unique_together = ("instance_id", "partition")
```
jamalex: Does this need to also include "profile"? Since we might have an empty partition for each profile for the same instance_id? Or do we need "profile" on the model at all?

morango/models.py
jamalex: Ah, this came in as I was commenting below.

> Since we serialize based on profile, I think this field is important for tracking that.

Since we don't expect partition collisions across profiles (except for empty string, for this model), we may be OK? In the empty string case, it doesn't matter which profile it came from.

ralphiee22: In the empty string case, it doesn't matter as long as we serialize all profiles?

jamalex: > In the empty string case, it doesn't matter as long as we serialize all profiles?

Even if we've only serialized one of multiple profiles, the existence of `(instance_id=X, partition="", counter=554)` in the DMC table means that we have _everything_ locally for that instance, up through that counter. As far as I'm able to reason it through, it shouldn't matter what profile the serializations happened under, as each one will increment that counter for the same instance_id.

jamalex: (Because there can't be a higher counter for another profile, since it all happened in this DB)

jamalex: (And if another profile hasn't been serialized in a long time, then the changes it has locally in the app layer aren't yet even associated with a counter, so they won't get skipped when we serialized them later, as they'll get new higher counters then)

PR #33, opened 2017-10-22
Synclient refactor

ralphiee22:
## Summary

Moved serializing, deserializing, queuing, and dequeuing to a utils file. Finished the synclient operations for querying certs, creating sync sessions, transfer sessions. Also updated APIs to perform the correct data operations when those endpoints are called.

## TODO

## Reviewer guidance


morango/api/viewsets.py
```diff
@@ -18,6 +18,7 @@ class CertificateViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.CertificateSerializer
     authentication_classes = (permissions.BasicMultiArgumentAuthentication,)
 
+    # @decorators.authentication_classes(permissions.BasicMultiArgumentAuthentication,)
```
ralphiee22: I was wondering whether we should only have auth permissions on creating a cert. And allow anyone to query the certs as in `get_remote_certificates`, which will get all certs the server owns.

morango/syncsession.py
```diff
@@ -173,334 +193,152 @@ def _close_sync_session(self, sync_session):
...
 
     def _close_transfer_session(self):
 
         # "delete" transfer session on server side
         self.sync_connection._close_transfer_session(self.current_transfer_session)
 
-        # delete local buffered objects if pushing records
-        if self.current_transfer_session.push:
-            Buffer.objects.filter(transfer_session=self.current_transfer_session).delete()
-            RecordMaxCounterBuffer.objects.filter(transfer_session=self.current_transfer_session).delete()
```
jamalex: Might there not be other times when we would call `_close_transfer_session`? E.g. if someone canceled? And we'd want to clear these out in that case also.

morango/utils/sync_utils.py
```diff
@@ -0,0 +1,440 @@
...
+        _dequeuing_merge_conflict_rmcb(cursor, transfersession.id)
+        _dequeuing_update_rmcs_last_saved_by(cursor, current_id, transfersession.id)
+        _dequeuing_delete_mc_rmcb(cursor, transfersession.id)
+        _dequeuing_delete_mc_buffer(cursor, transfersession.id)
+        _dequeuing_insert_remaining_buffer(cursor, transfersession.id)
+        _dequeuing_insert_remaining_rmcb(cursor, transfersession.id)
+        _dequeuing_delete_remaining_rmcb(cursor, transfersession.id)
+        _dequeuing_delete_remaining_buffer(cursor, transfersession.id)
+    if getattr(settings, 'DESERIALIZE_AFTER_DEQUEUING', True):
+        _deserialize_from_store(transfersession.sync_session.profile)
```
jamalex: Can you flag anything that changed in here, beyond just being ported over?

morango/syncsession.py
```diff
@@ -173,334 +193,152 @@ def _close_sync_session(self, sync_session):
...
+        self.sync_session.save()
+        self.sync_session = None
 
     def _create_transfer_session(self, push, filter):
 
         # build data for creating transfer session on server side
         data = {
             'id': uuid.uuid4().hex,
-            'filter': filter,
+            'filter': filter.__str__(),
```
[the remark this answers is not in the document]
ralphiee22: addressed in later commit.

morango/utils/sync_utils.py
```diff
@@ -207,11 +227,9 @@ def _queue_into_buffer(transfersession):
...
                             (instance_id, counter, transfer_session_id, model_uuid)
                             SELECT instance_id, counter, '{transfer_session_id}', store_model_id
-                            FROM {record_max_counter} AS rmc, {outgoing_buffer} AS buffer
-                            WHERE EXISTS (SELECT 1
-                                          FROM {outgoing_buffer}
-                                          WHERE buffer.model_uuid = rmc.store_model_id)
-                                          AND buffer.transfer_session_id = '{transfer_session_id}'
+                            FROM {record_max_counter} AS rmc LEFT JOIN {outgoing_buffer} AS buffer
+                            WHERE buffer.model_uuid = rmc.store_model_id
+                            AND buffer.transfer_session_id = '{transfer_session_id}'
```
ralphiee22: The old query took an incredibly long time. This new query passes tests and is way faster.

morango/utils/sync_utils.py
```diff
@@ -25,6 +25,21 @@ def _self_referential_fk(klass_model):
                 return f.attname
     return None
 
+def _fsic_queuing_calc(fsic1, fsic2):
+    fsic_copy = dict(fsic1)
+    # fsic_copy = dict(fsic2)
```
ralphiee22: need to remove this.

morango/utils/sync_utils.py
```diff
@@ -25,6 +25,21 @@ def _self_referential_fk(klass_model):
                 return f.attname
     return None
 
+def _fsic_queuing_calc(fsic1, fsic2):
```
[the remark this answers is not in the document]
jamalex: I've been playing with a one-line version of this that may be a bit more readable (that may be subjective, but the deletion of keys threw me at first in the current version):

```
return {instance: fsic2.get(instance, 0) for instance, counter in iteritems(fsic1) if fsic2.get(instance, 0) < counter}
```

I did some speed benchmarks, and it seems to depend. With a small number of keys (< 1000), your version is substantially faster. With a large number of keys, my one-liner is faster. When the two dictionaries are identical (or don't have many differences), my version seems to be faster (probably because you're starting with a full dict and then deleting everything in it one by one). The numbers we're talking about are small (it's a fast function in any case), and since it may be fairly common for the fsics to be largely the same (with legacy instance IDs sticking around but not changing), you might consider the one-liner if it doesn't feel too obtuse. Alternatively, here's a few-liner that performs similarly:

```
    fsic = {}
    for instance, counter in iteritems(fsic1):
        fsic2counter = fsic2.get(instance, 0)
        if fsic2counter < counter:
            fsic[instance] = fsic2counter
    return fsic
```

morango/utils/sync_utils.py
jamalex: Cool, can't see any issues! :racehorse:

morango/utils/sync_utils.py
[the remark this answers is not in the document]
jamalex: Probably good to throw a couple simple unit tests at this function, too.

jamalex:
Note that some of my review comments are hidden behind "Show outdated", because I focused on the substantive stuff, in commit 55a7eaa.

morango/utils/sync_utils.py
```diff
@@ -162,15 +177,20 @@ def _queue_into_buffer(transfersession):
...
-    fsics.update(remote_fsic)
+    server_fsic = json.loads(transfersession.server_fsic)
+    client_fsic = json.loads(transfersession.client_fsic)
+
+    if transfersession.push:
+        fsics = _fsic_queuing_calc(client_fsic, server_fsic)
+    else:
+        fsics = _fsic_queuing_calc(server_fsic, client_fsic)
+
+    # if fsics are identical, then there is nothing to queue
```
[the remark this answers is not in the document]
jamalex: (just in terms of the comment wording, I mean -- behavior is OK)

morango/api/viewsets.py
```diff
@@ -258,15 +261,25 @@ def create(self, request):
             "push": is_a_push,
             "records_total": request.data.get("records_total") if is_a_push else None,
             "sync_session": syncsession,
-            "local_fsic": '{}',
-            "remote_fsic": request.data.get('local_fsic') or '{}',
+            "client_fsic": request.data.get('client_fsic') or '{}',
+            "server_fsic": '{}',
```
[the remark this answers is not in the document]
ralphiee22: I have that code locally. Just wanted to finalize the order things happen in. Can include in this PR or future one.

morango/utils/sync_utils.py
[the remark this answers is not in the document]
ralphiee22: I have integration tests that actually test this functionality.

morango/syncsession.py
```diff
@@ -21,13 +21,23 @@ def _join_with_logical_operator(lst, operator):
     op = ") {operator} (".format(operator=operator)
     return "(({items}))".format(items=op.join(lst))
 
+def _get_server_ip(url):
+    try:
+        return socket.gethostbyname(url)
```
jamalex: I don't think this can take a URL, it needs to be just the host. Should probably pass in `hostname` here instead of url.

```
In [3]: socket.gethostbyname("http://google.com")
---------------------------------------------------------------------------
gaierror                                  Traceback (most recent call last)
/home/jamalex/<ipython-input-3-3f638f07b01c> in <module>()
----> 1 socket.gethostbyname("http://google.com")

gaierror: [Errno -2] Name or service not known

In [4]: socket.gethostbyname("google.com")
Out[4]: '216.58.216.14'
```

morango/syncsession.py
```diff
@@ -21,13 +21,23 @@ def _join_with_logical_operator(lst, operator):
...
+    except:
+        return ''
 
-def _get_ip(url):
+def _get_client_ip_for_server(server_host, server_port):
+    server_port = server_port if server_port else 80
```
jamalex: More idiomatic would be just to set the default with `server_port=80` in the params. Except I don't think we should allow defaults here, let's just always be explicit. (It's not often going to be 80, more commonly 8080 or whatnot).

Ah, but I see that `urlparse` returns `None` and doesn't know to set the default to 80 based on the protocol. However, one thing to note is that the default port depends on http vs https.

Perhaps don't set a default here, and then change the calling code to: `_get_client_ip_for_server(hostname, url.port or (80 if url.scheme == "http" else 443))` (perhaps splitting the port extraction out into its own line)

morango/api/viewsets.py
```diff
@@ -288,10 +289,12 @@ def perform_destroy(self, transfersession):
...
-            # update database max counters
-            DatabaseMaxCounter.update_fsics(json.loads(transfersession.client_fsic),
-                                            json.loads(transfersession.server_fsic),
-                                            certificates.Filter(transfersession.filter))
+            # update database max counters but use latest fsics on server
+            with transaction.atomic():
+                server_fsic = DatabaseMaxCounter.calculate_filter_max_counters(Filter(transfersession.filter))
+                DatabaseMaxCounter.update_fsics(json.loads(transfersession.client_fsic),
+                                                server_fsic,
+                                                certificates.Filter(transfersession.filter))
```
jamalex: Woohoo! Seems like it would be more DRY, safe, and encapsulated to move this logic directly inside `update_fsics`, though, unless I'm missing something? It should have access to everything it needs (basically just the filter), right?

morango/utils/sync_utils.py
```diff
@@ -26,19 +26,15 @@ def _self_referential_fk(klass_model):
...
-    return fsic_copy
+    """
+    We set the lower counter between two same instance ids.
+    If an instance_id exists in one fsic but not the other we want to give that counter a value of 0.
+
+    :param fsic1: dictionary containing (instance_id, counter) pairs
+    :param fsic2: dictionary containing (instance_id, counter) pairs
+    :return ``dict`` of fsics to be used in queueing the correct records to the buffer
+    """
+    return {instance: fsic2.get(instance, -1) for instance, counter in iteritems(fsic1) if fsic2.get(instance, -1) < counter}
```
jamalex: The first counter we set for an instance should be 1, right? So this can just be 0, not -1, can't it? (and that's what the docstring says, too)

morango/syncsession.py
ralphiee22: I actually do just pass in the host, but I should change the param name to reflect that, but it could also be an ip that we are passing in.

morango/api/viewsets.py
ralphiee22: yeah I did this because the way the tests are written, they break when i try putting that inside the function.

morango/utils/sync_utils.py
ralphiee22: right, its just tests broke so I opted of changing the logic a bit instead of trying to modify the tests.

morango/syncsession.py
ralphiee22: whoops take that back, I forgot to pass in hostname,

morango/utils/sync_utils.py
jamalex: If the behavior is correct with 0, but it breaks the tests, shouldn't we... fix the tests? When the universe is eventually destroyed, there's a good chance it will be caused by an off-by-one error. :)

morango/api/viewsets.py
jamalex: Ok, if it's hard to change those assumptions in the tests, could you perhaps just open a tech debt issue for now, and we can address it later?

morango/models.py
```diff
@@ -301,16 +301,18 @@ class Meta:
...
+    def update_fsics(cls, fsics, sync_filter):
+        internal_fsic = DatabaseMaxCounter.calculate_filter_max_counters(sync_filter)
         updated_fsic = {}
-        for key, value in iteritems(fsic1):
-            if key in fsic2:
+        for key, value in iteritems(fsics):
+            if key in internal_fsic:
                 # if same instance id, update fsic with larger value
-                if fsic1[key] > fsic2[key]:
-                    updated_fsic[key] = fsic2[key]
```
ralphiee22: I think this logic was wrong before, I believe fsic2 should have actually been fsic1.

morango/models.py
```diff
@@ -301,16 +301,18 @@ class Meta:
...
+                if fsics[key] > internal_fsic[key]:
+                    updated_fsic[key] = fsics[key]
```
jamalex: I'm still wrapping my head around this approach, and whether there might be any issues. Here's where my head's at:

1. When we calculate the `internal_fsic`, some of the values might come from superset filters (rather than the filter we're using here itself).
2. This means that we might skip setting a DMC record for the current filter even if one doesn't exist for some instance_id at all.
3. This is a slightly different way of thinking about the DMC than we'd previously discussed, but I'm having trouble thinking of any way it would actually negatively impact behavior. If anything, it just keeps the DMC size down and should help performance, by not storing redundant entries in the DMC. Only negative impact I can imagine is that it makes it harder to reason about the state of the DMC if something *does* go wrong.

Do you see what I mean? Can you think of any possible issues with skipping saving these entries, as long as there's a superset filter with a higher value for that instance?

ralphiee22: Hard to see what could go wrong. I like to think of examples.
If we are syncing individual users at a time for a facility. Then we have filter, instance_id pairs for each user individually. Then we sync the whole facility, and it basically makes those other filters useless as we have the superset filter with a higher counter.
If we sync the whole facility first, and then do an individual user sync, the individual user filter counter will be higher. But if we re-sync the whole facility then that counter will be higher making the subset filter useless.

morango/models.py
```diff
@@ -421,15 +422,16 @@ def serialize(self):
...
             if f.attname in getattr(self, '_internal_mptt_fields_not_to_serialize', '_internal_fields_not_to_serialize'):
                 continue
+            if f.attname in data:
+                continue
             data[f.attname] = f.value_from_object(self)
         return data
 
     @classmethod
-    def deserialize(cls, dict_model):
+    def deserialize(cls, dict_model, **kwargs):
```
jamalex: When do you project we would need to pass in `**kwargs`?

morango/utils/sync_utils.py
```diff
@@ -10,6 +11,10 @@
 from morango.models import Buffer, DatabaseMaxCounter, DeletedModels, InstanceIDModel, RecordMaxCounter, RecordMaxCounterBuffer, Store
 from morango.utils.register_models import _profile_models
 
+# custom JSON serialization class
+JSON_SERIALIZER_CLASS_PATH = getattr(settings, "MORANGO_JSON_SERIALIZER_CLASS", "morango.utils.encoders")
+JSON_SERIALIZER_MODULE = importlib.import_module(JSON_SERIALIZER_CLASS_PATH)
+JSON_SERIALIZER = getattr(JSON_SERIALIZER_MODULE, 'MorangoJSONEncoder')
```
jamalex: The typical pattern is to include the class name in the class path, like in test class specifications for running specific test classes. E.g., in cache backend settings: https://docs.djangoproject.com/en/2.0/ref/settings/#caches

jamalex: So the default value would be `morango.utils.encoders.MorangoJSONEncoder`. That way you could have one file with multiple possible encoder classes in it -- the class doesn't need to have a particular special name.

morango/models.py
```diff
@@ -407,11 +407,12 @@ def save(self, update_dirty_bit_to=True, *args, **kwargs):
             self._morango_dirty_bit = False
         super(SyncableModel, self).save(*args, **kwargs)
 
-    def serialize(self):
+    def serialize(self, data={}):
```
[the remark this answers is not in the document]
ralphiee22: from the parent class? atleast thats how i did it in https://github.com/learningequality/kolibri/pull/2496/files#diff-6efacfe268cfdb28150182ee233780aeR509
wasn't sure what the best pattern was to do this.

morango/models.py
ralphiee22: from the parent class? atleast thats how i did it in https://github.com/learningequality/kolibri/pull/2496/files#diff-6efacfe268cfdb28150182ee233780aeR509
wasn't sure what the best pattern was to do this.

morango/utils/sync_utils.py
ralphiee22: Any suggestions on how to do that?
The way I did it, is what I found online that worked for python 2 and python 3.

morango/models.py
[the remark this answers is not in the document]
ralphiee22: https://github.com/learningequality/kolibri/issues/2817

morango/models.py
ralphiee22: https://github.com/learningequality/kolibri/issues/2817

morango/utils/sync_utils.py
ralphiee22: https://github.com/learningequality/morango/issues/39

PR #40, opened 2017-12-05
Change macaddress to be hashable identifier.

ralphiee22:
## Summary

Rename `macadress` to `node_id` and make it a hashable id of `macaddress` with database id.
Also there is a fix for migrations that failed on postgres. datetime default was longer then the default length allowed for a char.

## Issues addressed

No longer store personal identifiable information through macaddress.


morango/migrations/0007_auto_20171018_1615.py
```diff
@@ -113,7 +113,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='syncsession',
             name='connection_kind',
-            field=models.CharField(choices=[('network', 'Network'), ('disk', 'Disk')], default=datetime.datetime(2017, 10, 18, 21, 15, 12, 858664, tzinfo=utc), max_length=10),
+            field=models.CharField(choices=[('network', 'Network'), ('disk', 'Disk')], default='', max_length=10),
```
jamalex: Generally very unrecommended to modify an existing migration. But the migration hasn't been shipped anywhere previously, has it? In Kolibri?

jamalex: (and changing a default like this shouldn't be a big issue)

ralphiee22: yes to the 1st question and yeah thats what I thought about the default as well.

PR #44, opened 2018-01-18
Added endpoint which shows information about morango and the device.

ralphiee22:
## Summary

Added endpoint which will give a proquint hash based on the morango `instance_id`. Also displays `instance_id`, device OS, and version of morango. The proquint hash will change if the `instance_id` of the device changes.

## TODO

## Reviewer guidance

### DIFF
--- a/morango/models.py
+++ b/morango/models.py
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 
-import json
 import hashlib
+import json
 import os
 import platform
 import sys
@@ -15,8 +15,9 @@ from django.utils import timezone
 from django.utils.six import iteritems
 from morango.utils.register_models import _profile_models
 
-from .certificates import Certificate, ScopeDefinition, Nonce, Filter
+from .certificates import Certificate, Filter, Nonce, ScopeDefinition
 from .manager import SyncableModelManager
+from .utils import proquint
 from .utils.uuids import UUIDField, UUIDModelMixin, sha2_uuid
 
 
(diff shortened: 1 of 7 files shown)


morango/models.py
```diff
@@ -123,6 +137,11 @@ def get_or_create_current_instance(cls):
             obj.current = True
             obj.save()
 
+            MorangoInfo.objects.update_or_create(pk=1, defaults={'instance_hash': proquint.from_int(int(obj.id[:8], 16)),
+                                                                 'instance': obj,
+                                                                 'system_os': platform.system(),
+                                                                 'version': morango.__version__})
```
ralphiee22: Not sure if this is useful, or this is how we want to display this OS info to the user.

jamalex: Seems good for OS, yeah.

morango/models.py
```diff
@@ -14,12 +15,25 @@
...
 from django.utils.six import iteritems
 from morango.utils.register_models import _profile_models
+from .utils import proquint
 
 from .certificates import Certificate, ScopeDefinition, Nonce, Filter
 from .manager import SyncableModelManager
 from .utils.uuids import UUIDField, UUIDModelMixin, sha2_uuid
 
 
+class MorangoInfo(models.Model):
```
jamalex: I don't think we need a model for this. It should be simpler and cleaner (and still fast, as this stuff is cheap) to use a base `ViewSet`, define the `retrieve` method, and just compute/return the data directly.

For the proquint, perhaps that logic can be embedded in a `get_proquint` method on the `InstanceIDModel`, and called from the viewset.

Issue #37, opened 2017-12-04
Compatibility possible with cryptography 1.2+ ?


benjaoming:
As a test matrix supporting several versions of cryptography has been built, we could work on this in order to support native system libraries.

Ubuntu 16.04: cryptography 1.2.3

Alternatively / supplementary: We shouldn't allow versions that aren't supported, which is part of the error in https://github.com/learningequality/kolibri/issues/2805

One suggestion would be to create a check that also takes the version into account. Current check does this:

```
>>> try:
...     from cryptography.hazmat.backends import default_backend
...     from cryptography import exceptions as crypto_exceptions
...     crypto_backend = default_backend()
...     from cryptography.hazmat.primitives.asymmetric import rsa as crypto_rsa, padding as crypto_padding
...     from cryptography.hazmat.primitives import serialization as crypto_serialization, hashes as crypto_hashes
...     CRYPTOGRAPHY_EXISTS = True
... except:
...     CRYPTOGRAPHY_EXISTS = False
```

benjaoming:
Cryptography 1.2 cannot sign stuff with a private key, so it can never be supported and must just be ignored as implemented in #46

PR #32, opened 2017-10-20
Expand test matrix w/ different django+cryptography releases

benjaoming:
## Summary

Changes tox to use current Python environment and sets Travis to test with Python 2.7, 3.4, 3.5, 3.6

PR #46, opened 2018-01-28
Cryptography 1.2.3 support

benjaoming:
## Summary

Trying to see if I can implement support for older cryptography versions so they don't mess up when they're present on the host system.

## TODO

## Reviewer guidance

In progress...

## Issues addressed

#32 #37 

## Documentation

n/a

### DIFF
--- a/morango/crypto.py
+++ b/morango/crypto.py
@@ -18,6 +18,9 @@ try:
     crypto_backend = default_backend()
     from cryptography.hazmat.primitives.asymmetric import rsa as crypto_rsa, padding as crypto_padding
     from cryptography.hazmat.primitives import serialization as crypto_serialization, hashes as crypto_hashes
+    # Ignore cryptography versions that do not support the 'sign' method
+    if not hasattr(crypto_rsa.RSAPrivateKey, 'sign'):
+        raise ImportError
     CRYPTOGRAPHY_EXISTS = True
 except:
     CRYPTOGRAPHY_EXISTS = False
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,5 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+tests/testapp/testapp.db
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,11 @@ python:
 env:
 # Removed because linting PR coming up separately
 #  - LINT="yes"
+  - DJANGO="1.9" CRYPTOGRAPHY="1.2.3"
   - DJANGO="1.9" CRYPTOGRAPHY="2.0"
+  - DJANGO="1.10" CRYPTOGRAPHY="1.2.3"
   - DJANGO="1.10" CRYPTOGRAPHY="2.0"
+  - DJANGO="1.11" CRYPTOGRAPHY="1.2.3"
   - DJANGO="1.11" CRYPTOGRAPHY="2.0"
 
 matrix:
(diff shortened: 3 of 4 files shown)


benjaoming:
Nah, seems impossible? I cannot find any implementation in cryptography 1.2.3 that supports private key signing of data?

So the ideal would be if Morango ignores cryptography 1.2.3...?

benjaoming:
Very broken on 1.2.3 :)

```
tests/testapp/tests/test_api.py FFFFFFFFFFF...FFFFFFFFFFFFFFFFFFFFFFFFFFFFF.                                                                                               [ 23%]
tests/testapp/tests/test_certificates.py FFFFFFFFFF.                                                                                                                       [ 28%]
tests/testapp/tests/test_connection.py FFFFF.....                                                        
```

benjaoming:
With tests clocking in at 19+ minutes, this clearly needs a very scoped test matrix :) @jamalex @ralphiee22 any opinions?

morango/crypto.py
```diff
@@ -18,6 +18,9 @@
     crypto_backend = default_backend()
     from cryptography.hazmat.primitives.asymmetric import rsa as crypto_rsa, padding as crypto_padding
     from cryptography.hazmat.primitives import serialization as crypto_serialization, hashes as crypto_hashes
+    # Ignore cryptography versions that do not support the 'sign' method
+    if not hasattr(crypto_rsa.RSAPrivateKey, 'sign'):
+        raise ImportError
```
jamalex: Interesting that it didn't support this -- seems like a pretty fundamental operation!

Seems like a reasonable approach here. We should perhaps also specify a minversion in the requirements?

jamalex:
We should probably change the requirements to be `>=` rather than `==`?
https://github.com/learningequality/morango/blob/master/requirements/accelerated.txt#L3

And perhaps then also add 1.8.0 to the test matrix?

ralphiee22:
Curious why we want to add 1.8 to the test matrix? b/c it is a LTS version?

ralphiee22:
if we are doing `>=` for the cryptography requirements, can we take out `1.2.3` from the test matrix?

benjaoming:
@ralphiee22 the idea to add `cryptography` to the test matrix was mainly to say that "if present" it's not actually used, but that the tests succeed.

Seeing that they run so incredibly slowly though, I'd prefer to run the tests only for Django 1.11 (currently the Kolibri 0.8 version, and probably going to be the final and only Django version because of Python 2.7)

benjaoming:
I've updated the test matrix such that cryptography 1.2.3 is tested only for Django 1.11 + Python 2.7 and Django 1.11 + Python 3.6

benjaoming:
Also added 1.8.1 with tests for all Python + Django versions (since it's accellerated)

benjaoming:
Sorry about adding such a slow test, could consider to expand the test matrix with build stages: https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions

benjaoming:
I cannot grasp the stage rules at this point... am confused about how to include/exclude versions of Python in the matrix.. will just open an issue about it.

Kolibri would benefit a lot from such staged test behavior as well, to separate tests into slow quick-fail tests before running the full set.

jamalex:
Not too worried about morango travis speed at this stage, as we can run the primary test set locally during development. We don't open as many Morango PRs as we do on Kolibri, so it shouldn't be a big slowdown to our flow.

benjaoming:
The latest commit has already been queued for 1 hour.. I think maybe Travis has some way of down-prioritizing these long, slow builds.

benjaoming:
Anyways, if the tests succeed, I think it can be released? @ralphiee22 ?

Ping me if this is released, and we can propose the version bump for Kolibri 0.7.1 immediately :)

PR #47, opened 2018-02-10
Halt tox when model changes are not reflected in migrations

benjaoming:
## Summary

We need to fail if migrations are missing, as they seem to be in kolibri 0.7.0 through to 0.7.2-beta2

## TODO

## Reviewer guidance

n/a

## Issues addressed

Adressing https://github.com/learningequality/kolibri/issues/3162 and https://github.com/learningequality/kolibri/pull/3180

## Documentation

n/a


benjaoming:
Notice that this did not spawn any errors, but that's because tests already installed `django-mptt==0.8.7` in which the migration specifying a default manager is avoided, also in Django 1.9. Proof of this is found in https://github.com/learningequality/kolibri/pull/3180

benjaoming:
Cryptography 1.8 builds are now clocking in at ~2 minutes, not ~20. This was a mistake. There are only 2 builds with Cryptography 1.2.3, which are expected to take ~20 minutes.

Run time had clearly blown up :)

image

benjaoming:
Phew! Sorry for the wasted computation resources, Travis

image

PR #61, opened 2018-10-09
Hard deletion

ralphiee22:
## Summary

Adds the ability to delete a model and its identifying data from the store. The deletion will propagate to other devices when syncing.

## TODO

## Issues addressed

GDPR compliant.
https://github.com/learningequality/kolibri/issues/4303


morango/models.py
```diff
@@ -437,6 +452,24 @@ def save(self, update_dirty_bit_to=True, *args, **kwargs):
             self._morango_dirty_bit = False
         super(SyncableModel, self).save(*args, **kwargs)
 
+    def delete(self, using=None, keep_parents=False, hard_delete=False, *args, **kwargs):
```
ralphiee22: A lot of this logic is taken from django's standard `delete` method. This adds the tracking of IDs of the related models to also hard delete.

morango/models.py
```diff
@@ -288,17 +288,33 @@ def _deserialize_store_model(self):
         klass_model = _profile_models[self.profile][self.model_name]
         # if store model marked as deleted, attempt to delete in app layer
         if self.deleted:
-            klass_model.objects.filter(id=self.id).delete()
+            # if hard deleted, propogate to related models
```
jamalex: *propagate

morango/models.py
```diff
@@ -288,17 +288,33 @@ def _deserialize_store_model(self):
...
             app_model._morango_partition = self.partition
             try:
+                app_model.clean_fields()
                 with mute_signals(signals.pre_save, signals.post_save):
                     app_model.save(update_dirty_bit_to=False)
             # if unable to save due to missing FKs, mark model as deleted
-            except ObjectDoesNotExist:
+            except ValidationError:
+                # check FKs in store to see if any of those models were hard deleted
+                fk_ids = [getattr(app_model, field.attname) for field in app_model._meta.fields if isinstance(field, ForeignKey)]
```
jamalex: I suppose it might be good to include OneToOneFields as well? Oh, nevermind, they're a subclass of ForeignKey! :fireworks:

morango/models.py
```diff
@@ -288,17 +288,33 @@ def _deserialize_store_model(self):
...
+                for fk_id in fk_ids:
+                    try:
+                        if Store.objects.get(id=fk_id).hard_delete:
+                            app_model._update_hard_deleted_models()
+                    except Store.DoesNotExist:
+                        pass
                 app_model._update_deleted_models()
```
jamalex: As discussed in a comment on the commit -- probably best to only do the auto-deletion when we explicitly see that the FK points to a deleted record in the store (can do jointly with what you do for hard-deletion, just above this). Otherwise, leaving the dirty bit set so it can try again later.

jamalex:
Cool! Added a couple notes on that commit. Another quick question: have we checked that we can recreate a record (same id) that was previously deleted? How about hard-deleted? Even if it works now, might be good to add explicit test coverage for this cases if we haven't.

PR #63, opened 2018-11-05
create dict mapping store models to reduce reads

ralphiee22:
## Summary

We use `in_bulk` which creates a dict mapping id -> model. This allows us to read all store instances per syncable model class. 

## Issues addressed

Reduces having to read each store model separately (for already serialized models).


morango/utils/sync_utils.py
```diff
@@ -62,9 +62,10 @@ def _serialize_into_store(profile, filter=None):
             klass_queryset = klass_model.objects.filter(_morango_dirty_bit=True)
             if prefix_condition:
                 klass_queryset = klass_queryset.filter(prefix_condition)
+            store_records_dict = Store.objects.in_bulk(id_list=klass_queryset.values_list('id', flat=True))
```
jamalex: Oooh, fancy, didn't know this one!

PR #64, opened 2018-11-08
Bypass DRF validation but still do security checks

ralphiee22:
## Summary

Moved the security check validation code from the serializer to the viewset. Also, `bulk_create` the buffer and recordmaxcounter buffer records.

## Reviewer guidance


tests/testapp/tests/test_api.py
```diff
@@ -747,7 +747,7 @@ def test_push_fails_for_pull_transfersession(self):
         rec_1 = self.build_buffer_item(push=False, filter=self.default_push_filter)
         rec_2 = self.build_buffer_item(transfer_session=rec_1.transfer_session)
         rec_3 = self.build_buffer_item(transfer_session=rec_1.transfer_session)
-        self.make_buffer_post_request([rec_1, rec_2, rec_3], expected_status=400)
+        self.make_buffer_post_request([rec_1, rec_2, rec_3], expected_status=403)
```
ralphiee22: This test was actually failing based on non valid sub partitions and not on push/pull transfer session mismatch. It now fails in the correct spot and returns the correct error code.

morango/api/viewsets.py
```diff
@@ -375,30 +377,59 @@ class BufferViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
...
+            try:
+                Model = _profile_models[record["profile"]][record["model_name"]]
+            except KeyError:
+                Model = SyncableModel
+
+            partition = record['partition'].replace(record['model_uuid'], Model.ID_PLACEHOLDER)
+            expected_model_uuid = Model.compute_namespaced_id(partition, record["source_id"], record["model_name"])
+            if expected_model_uuid != record["model_uuid"]:
+                return response.Response({"model_uuid": "Does not match results of calling {}.compute_namespaced_id".format(Model.__class__.__name__)}, status=status.HTTP_400_BAD_REQUEST)
+
```
jamalex: Looks like there's an old version of the validation logic left in the serializer? https://github.com/learningequality/morango/blob/7853d2142dd266bfc06a8f673899549f8456c4e8/morango/api/serializers.py#L93

Perhaps good to clear all the unused stuff out over there to avoid confusion?

morango/validation.py
```diff
@@ -0,0 +1,49 @@
...
+        # ensure the provided model_uuid matches the expected/computed id
+        try:
+            Model = _profile_models[record["profile"]][record["model_name"]]
+        except KeyError:
+            Model = SyncableModel
+
+        partition = record['partition'].replace(record['model_uuid'], Model.ID_PLACEHOLDER)
+        expected_model_uuid = Model.compute_namespaced_id(partition, record["source_id"], record["model_name"])
+        if expected_model_uuid != record["model_uuid"]:
+            return {"model_uuid": "Does not match results of calling {}.compute_namespaced_id".format(Model.__class__.__name__)}
```
jamalex: Raise exception instead

PR #65, opened 2018-11-10
Cache foreign key lookups across models

ralphiee22:
## Summary

As we call `clean_fields` on each model, this does a DB read for foreign keys. This caches same object foreign key look ups between models. Speeds up the deserialization process.


morango/models.py
```diff
@@ -437,6 +439,42 @@ def save(self, update_dirty_bit_to=True, *args, **kwargs):
             self._morango_dirty_bit = False
         super(SyncableModel, self).save(*args, **kwargs)
 
+    def clean_fields(self, exclude=None):
```
jamalex: I'm trying to think through whether overriding `clean_fields` entirely might cause issues elsewhere it could be used. Perhaps we might do the FK-checking piece in a separate function (that we call during deserialization), and then have that function call the real `clean_fields` passing `exclude=[thefkfields]` so it doesn't redo those ones?

Our special function could then also just be passed a shared dict for the caching, so it can easily be tossed at the end of the deserialization -- I hadn't thought about the risks of the cache persisting beyond the transaction.

Can chat this afternoon if you want!

ralphiee22: I would opt for this approach. I was looking for something that we can only use during deserialization. This is a good workaround.

PR #66, opened 2018-12-18
bulk deserialize sql query

ralphiee22:
## Summary

Load up the data from each app model on a per class basis, and insert into the database with a single query. This strategy is only used on app models which don't inherit from mptt. Different implementations depending on db backend. For Postgres, we have to cast the fields to their appropriate types or else the query errors out.

**Note**: This PR passes the tests mentioned here https://github.com/learningequality/kolibri/blob/develop/kolibri/core/auth/test/test_morango_integration.py
## TODO


morango/models.py
```diff
@@ -289,9 +290,9 @@ def _deserialize_store_model(self):
         When deserializing a store model, we look at the deleted flags to know if we should delete the app model.
         Upon loading the app model in memory we validate the app models fields, if any errors occurs we follow
         foreign key relationships to see if the related model has been deleted to propagate that deletion to the target app model.
-        If there are no errors, we save the app model.
-
-        :return: ``bool`` whether the object was successfully deleted or saved
+        We return:
+        (None, True/False) => if the model was deleted successfully or it did not validate
+        (model, True) => if the model validates successfully
```
jamalex: This framing is a bit confusing to me. Perhaps separately explain each piece of the tuple, or... something?

morango/models.py
```diff
@@ -305,21 +306,21 @@ def _deserialize_store_model(self):
...
+                # validate the model
                 app_model.clean_fields()
-                with mute_signals(signals.pre_save, signals.post_save):
-                    app_model.save(update_dirty_bit_to=False)
-                return valid
-            except exceptions.ValidationError:
+                return app_model, valid
+            except exceptions.ValidationError as e:
                 valid = False
+                logger.debug("Validation error for {model} with id {id}: {error}".format(model=klass_model.__name__, id=app_model.id, error=e))
```
jamalex: Might be good to at least have this as a `logger.warn`, so it will show up in production logs people might share with us.

morango/models.py
jamalex: Or, if you feel it simplifies it a bit, perhaps switch to Django-style validation where errors trigger an Exception, and then you just need a single return value.

morango/utils/backends/sqlite.py
```diff
@@ -6,6 +6,14 @@
 class SQLWrapper(BaseSQLWrapper):
     backend = 'sqlite'
 
+    def _bulk_insert_into_app_models(self, cursor, app_model, fields, db_values, placeholder_str):
+        fields = str(tuple(str(f.attname) for f in fields)).replace("'", '')
+        insert = """REPLACE INTO {app_model} {fields}
+                    VALUES {placeholder_str}
+        """.format(app_model=app_model, fields=fields, placeholder_str=placeholder_str)
```
jamalex: Wow, so much simpler than Postgres! :ocean:

morango/utils/backends/postgres.py
```diff
@@ -2,10 +2,46 @@
...
 class SQLWrapper(BaseSQLWrapper):
     backend = 'postgresql'
 
+    def _bulk_insert_into_app_models(self, cursor, app_model, fields, db_values, placeholder_str):
+        # cast the values in the SET statement to their appropiate postgres db types
+        set_casted_values = ', '.join(map(lambda f: '{f} = nv.{f}::{type}'.format(f=f.attname, type=f.rel_db_type(connection)), fields))
+        # cast the values in the SELECT statement to their appropiate posgtres db types
+        select_casted_values = ', '.join(map(lambda f: '{f}::{type}'.format(f=f.attname, type=f.rel_db_type(connection)), fields))
+        # cast the pk to a UUID type from a Char type so postgres doesn't complain
+        pk = [f for f in fields if f.primary_key][0]
```
jamalex: What if there are other UUID fields? And other special types? Do we not more generally need to use something like `Field.prepare_value` or whatever it is to do proper conversion? Feels like this could be brittle.

ralphiee22: sorry the comment may have been a bit misleading since it specifically solves the problem for kolibri, but below we find the proper type
https://github.com/learningequality/morango/pull/66/files/8c6d941fd1e3d9a3ae6a3a894183ebb974994010#diff-9d55d7ca123fd15267bc3f96034b1380R41

morango/models.py
rtibbles: Yeah, I think throwing an exception for validation failure, and then just returning model or None, depending on whether it was deleted or not makes the most sense.

morango/utils/backends/postgres.py
rtibbles: Yeah, definitely worth updating the comment, as it makes it sound like you are casting all PKs to UUIDs.

morango/utils/sync_utils.py
```diff
@@ -171,8 +174,11 @@ def _deserialize_from_store(profile):
                 # keep iterating until size of dirty_children is 0
                 while len(dirty_children) > 0:
                     for store_model in dirty_children:
-
-                        if store_model._deserialize_store_model():
+                        (app_model, valid) = store_model._deserialize_store_model()
```
rtibbles: Can just do a try except here for validation.

PR #69, opened 2019-05-19
pwd is not available in Windows

jredrejo:
## Summary

Latest changes in file.py introduced a bug in Windows because it tries to use the `pwd` module , which is unix only. 

## Reviewer guidance

It's a simple check

## Issues addressed

Ensure a correct writable path is used

### DIFF
--- a/morango/constants/file.py
+++ b/morango/constants/file.py
@@ -1,6 +1,4 @@
-import pwd
 import os
 
-SQLITE_VARIABLE_FILE_CACHE = os.path.join(pwd.getpwuid(os.getuid()).pw_dir,
+SQLITE_VARIABLE_FILE_CACHE = os.path.join(os.path.expanduser('~'),
                                           'SQLITE_MAX_VARIABLE_NUMBER.cache')
-
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,5 +1,5 @@
 django<1.12
-django-mptt>=0.8.0
+django-mptt<0.10.0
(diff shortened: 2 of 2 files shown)


rtibbles:
@ralphiee22 this PR looks fine, but there appears to be pre-existing issue of unrecognized migrations under Python3. Migrations should be py2/3 compatible.

ralphiee22:
The migration it wants to make is to remove the DB index from the mptt fields `lft`, `level`, `rght`. Unsure why this is an issue now.

kollivier:
Taking a quick look, the failing build is referencing a 0013 migration, which does not appear to be in master or in this PR. One thing to note was that about a month ago mptt released a new version, that doesn't even have release notes, so hard to say but maybe it made some change that is causing the CI builds to create a new migration?

ralphiee22:
I think you are right @kollivier. I thought mptt was pinned to a specific version, but its actually `<=`, so that's probably what it is.

rtibbles:
Ah, that would kind of explain it - still curious why the differential between 2 and 3 though?

ralphiee22:
I updated this PR, but I am still wondering why this hasn't affected the migrations within kolibri?

rtibbles:
Because we explicitly pin django-mptt in Kolibri: https://github.com/learningequality/kolibri/blob/develop/requirements/base.txt#L9

ralphiee22:
I see @kollivier is referring to `0.10.0` in django mptt now.

PR #70, opened 2019-05-29
Add gzipping and chunking of data

ralphiee22:
## Summary

Adds parser to be used when the content-type in header is set to gzip. We only gzip if the client and server both have gzip capability. Also adds parameter to specify chunking of buffer requests.

## TODO


morango/syncsession.py
```diff
@@ -95,11 +116,14 @@ def _request(self, endpoint, method="GET", lookup=None, data={}, params={}, user
             lookup = lookup + '/'
         url = urljoin(urljoin(self.base_url, endpoint), lookup)
         auth = (userargs, password) if userargs else None
-        resp = requests.request(method, url, json=data, params=params, auth=auth)
+        if gzip:
```
ralphiee22: Given the way things are structured, found it easiest to separate into 2 requests.

morango/parsers.py
```diff
@@ -0,0 +1,18 @@
+from rest_framework.parsers import BaseParser
+import gzip
```
jamalex: Probably best to put this import inside the `parse` method, as importing it at this level could already throw the exception. Since this module isn't specific to gzip, we would want to be able to import from this file without errors.

morango/parsers.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    media_type = 'application/gzip'
+
+    def parse(self, stream, media_type=None, parser_context=None):
+        """
+        Parses the incoming bytestream by decompressing the gzipped data and returns the resulting data as a dictionary.
+        """
+        with gzip.GzipFile(fileobj=io.BytesIO(stream.read())) as f:
+            data = f.read()
+        return json.loads(data.decode('utf-8'))
```
jamalex: Do your tests include Unicode text?

morango/syncsession.py
```diff
@@ -55,6 +61,14 @@ def _get_client_ip_for_server(server_host, server_port):
     return IP
 
 
+# Used from https://github.com/django/django/blob/1.11.20/django/utils/text.py#L295
```
jamalex: Might be clearer as "Borrowed from..." -- I was trying to figure out why Django would be importing from Morango! :)

morango/syncsession.py
```diff
@@ -70,10 +84,17 @@ class Connection(object):
...
         self.base_url = base_url
-
-    def _request(self, endpoint, method="GET", lookup=None, data={}, params={}, userargs=None, password=None):
+        self.compresslevel = compresslevel
+        # ping server at url with info request
+        info_url = urljoin(self.base_url, api_urls.INFO)
+        self.server_info = requests.get(info_url).json()
+        self.capabilities = set()
+        if self.server_info.get('capabilities'):
+            self.capabilities = CAPABILITIES.intersection(set(self.server_info['capabilities']))
```
jamalex: If you want, could probably simplify down these 3 lines into:
`self.capabilities = CAPABILITIES.intersection(self.server_info.get('capabilities', []))`

morango/syncsession.py
```diff
@@ -70,10 +84,17 @@ class Connection(object):
...
+
+    def _request(self, endpoint, method="GET", lookup=None, data={}, params={}, userargs=None, password=None, gzip=False):
```
jamalex: A param called `gzip` suggests to me it's going to do the gzipping inside here, but if I understand it correctly, this is more something like `data_is_gzipped`?

morango/syncsession.py
```diff
@@ -278,7 +302,13 @@ def _close_sync_session(self, sync_session):
 
     def _push_record_chunk(self, serialized_recs):
         # push a chunk of records to the server side
-        return self._request(api_urls.BUFFER, method="POST", data=serialized_recs)
+        data = serialized_recs
+        GZIP = GZIP_BUFFER_POST in self.capabilities
```
jamalex: Suggest `use_gzip` (more semantic wording, and lowercase because it's not a constant)

tests/testapp/tests/test_api.py
```diff
@@ -716,17 +717,27 @@ def build_buffer_item(self, **kwargs):
 
         return buffermodel
 
-    def make_buffer_post_request(self, buffers, expected_status=201):
+    def make_buffer_post_request(self, buffers, expected_status=201, gzip=False):
```
jamalex: This name in this case makes sense.

PR #73, opened 2019-06-13
Switch from file-based caching to Django's lru_cache for sqlite max vars

jamalex:
There were several issues arising in Kolibri as a result of the pesky `SQLITE_MAX_VARIABLE_NUMBER.cache` file, so this PR switches to just using Django's `lru_cache` to memoize the result of the function, so we can call it as many times as we want. It's not so expensive to compute the first time (70ms for me), and only needs to be run once per process, so I think this should be fine.

### DIFF
--- a/morango/__init__.py
+++ b/morango/__init__.py
@@ -3,4 +3,4 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 default_app_config = "morango.apps.MorangoConfig"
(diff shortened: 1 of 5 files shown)


jamalex:
Note: released to PyPI as 0.4.6 for integration into Kolibri, but will wait to merge here until it's been reviewed

PR #78, opened 2020-02-18
allow for session metadata through settings config

ralphiee22:
## Summary

Add metadata about "this" device for each sync session through a `MORANGO_METADATA` settings attribute.


ralphiee22:
I apologize for the spam here. I was testing some webhook events.

indirectlylit:
looks like it worked! :)

> image

PR #81, opened 2020-05-08
Dependency fixes and cleanup

bjester:
## Summary
The existing `setup.py` was relying on pip's internal parsing tools to parse requirements from the `requirements.txt`. It would seem that pip has [recently updated](https://github.com/pypa/pip/issues/8188#issuecomment-623428010) that such that the existing implementation no longer works on newer versions of pip. This update moves the production dependencies to the `setup.py` and updates the other requirements files and tox appropriately. The pinned version of tox also needed updating.

## TODO

- [ ] ~~Have tests been written for the new code?~~
- [ ] ~~New dependencies (if any) added to requirements file~~

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -11,6 +11,6 @@ Please feel free to add your name to this list if you make a PR
 * Rafael Aguayo (ralphiee22)
 * Jamie Alexandre (jamalex)
 * Benjamin Bach (benjaoming)
-* Jose Rodriguez (jredrejo)
+* José Redrejo Rodríguez (jredrejo)
 * Aron Asor (aronasorman)
 * Blaine Jester (bjester)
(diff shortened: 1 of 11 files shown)


AUTHORS.md
```diff
@@ -13,3 +13,4 @@ Please feel free to add your name to this list if you make a PR
 * Benjamin Bach (benjaoming)
 * Jose Rodriguez (jredrejo)
```
jredrejo, line 14: Blaine, if you don't mind fixing this:Someone wrote my name incorrectly, It's José Redrejo (or José Redrejo Rodríguez, but not José Rodríguez :) )

bjester: @jredrejo Will do!

PR #82, opened 2020-05-12
Sync signals and bandwidth tracking

bjester:
## Summary

This adds tracking to the requests and responses during sync. The tracking sums the content length of each request and response, so the consumer of the `SyncClient` can be informed of the data transfer while it's syncing. In order to then notify the consumer of the client, signaling was added so the consumer can hook into queuing, dequeuing, pushing, and pulling stages of the sync process.

## TODO

- [ ] ~New dependencies (if any) added to requirements file~

## Reviewer guidance

[Prerequisite PR](https://github.com/learningequality/morango/pull/81)
[Notion project](https://www.notion.so/learningequality/Backend-Admins-can-sync-data-between-peers-on-LAN-8-13-pts-1f89f15bd98045ee86b7e3b6203fac39)


morango/sync/syncsession.py
```diff
@@ -423,25 +508,21 @@ def initiate_push(self, sync_filter):
...
-        logger.info("Beginning pulling of data...")
-        self._pull_records()
-        logger.info("Completed pull of data")
-        _dequeue_into_store(self.current_transfer_session)
+        with self.pulling.send(
+            transfer_session=self.current_transfer_session
+        ) as status:
+            self._pull_records(status.in_progress.fire)
+
+        # raise RuntimeError("oops")
```
bjester, line 522: Oops

bjester:
Hey @jamalex could you take a look at this?

morango/sync/syncsession.py
```diff
@@ -423,25 +528,19 @@ def initiate_push(self, sync_filter):
         self._close_transfer_session()
 
     def initiate_pull(self, sync_filter):
-
-        logger.info("Initiating pull sync")
         self._create_transfer_session(False, sync_filter)
 
         if self.current_transfer_session.records_total == 0:
-            logger.info("There are no records to transfer")
```
jamalex, line 431: This may still be a useful logger output to keep.

jamalex: But may make sense to shift this block (lines 533-535) down into the signal context (before line 540), so that the signals would still be fired (even if in rapid succession), in case something external was expecting them.

morango/sync/syncsession.py
```diff
@@ -423,25 +528,19 @@ def initiate_push(self, sync_filter):
...
             self._close_transfer_session()
             return
 
-        logger.info(
-            "{} records have been queued server side for transfer".format(
-                self.current_transfer_session.records_total
-            )
-        )
-        # pull records and close transfer session upon completion
-        logger.info("Beginning pulling of data...")
```
jamalex, line 441: I couldn't find where these specific ones were replaced by use of `OperationLogger` -- it'd be helpful to still print these for when we're running from the command line. Same for pushing.

morango/sync/syncsession.py
```diff
@@ -423,25 +528,19 @@ def initiate_push(self, sync_filter):
...
-        self._pull_records()
-        logger.info("Completed pull of data")
-        _dequeue_into_store(self.current_transfer_session)
+        with self.pulling.send(
+            transfer_session=self.current_transfer_session
+        ) as status:
+            self._pull_records(status.in_progress.fire)
```
jamalex, line 540: Since it's a "named argument" (in the sense that it has a default), may be better to explicitly say `callback=status.in_progress.fire` here.

morango/sync/syncsession.py
```diff
@@ -485,20 +584,22 @@ def _pull_records(self, callback=None):
                     "Specified TransferSession does not match this SyncClient's current TransferSession."
                 )
 
-            validate_and_create_buffer_data(data, self.current_transfer_session)
+            validate_and_create_buffer_data(
+                data, self.current_transfer_session, self.sync_connection
```
jamalex, line 588: `connection=self.sync_connection` could be clearer here, since it's an optional/named argument

morango/sync/syncsession.py
```diff
@@ -572,9 +672,18 @@ def _create_transfer_session(self, push, filter):
...
         self.current_transfer_session.save()
+
+        # the next call to create the session starts queueing on server side
+        if not push:
+            self.queuing.started.fire(local=False)
+
         # create transfer session on server side
         transfer_resp = self.sync_connection._create_transfer_session(data)
 
+        # the next call to create the session starts queueing on server side
```
jamalex, line 683: May need to update this comment (cloned from above?)

morango/proquint.py
```diff
@@ -58,7 +59,7 @@ def from_int(data):
     :type data: int
     :rtype: string
     """
-    if not isinstance(data, int) and not isinstance(data, long):
+    if not isinstance(data, int) and not isinstance(data, six.integer_types):
```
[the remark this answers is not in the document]
bjester: Yes, my bad I didn't notice that. Thanks!

morango/sync/syncsession.py
```diff
@@ -376,42 +384,139 @@ def _pull_record_chunk(self, chunk_size, transfer_session):
...
+        self.completed.fire()
+
+
 class SyncClient(object):
     """
     Controller to support client in initiating syncing and performing related operations.
     """
 
+    session = SyncSignalGroup()
+    """Signal group firing for each push and pull `TranferSession`."""
```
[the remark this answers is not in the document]
bjester: Yes it gets picked up for the preceding attribute. It's the same procedure as you would put a comment after defining the class or a function name.

morango/sync/syncsession.py
bjester: With the signals, the consumer of the morango API can now implement their own logging. This avoids unnecessary noise.

morango/sync/syncsession.py
```diff
@@ -376,42 +384,139 @@ def _pull_record_chunk(self, chunk_size, transfer_session):
...
         self.chunk_size = chunk_size
         self.current_transfer_session = None
 
     def initiate_push(self, sync_filter):
-
-        logger.info("Initiating push sync")
         self._create_transfer_session(True, sync_filter)
 
-        _queue_into_buffer(self.current_transfer_session)
+        with self.queuing.send(local=True):
```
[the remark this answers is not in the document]
bjester: I did consider passing the transfer session to all signals, since we could infer local/remote based off that object. It wasn't required for the implementation so I avoided adding it. 

I'm not sure I understand how multiple syncs would occur for the client. If you're talking about a push AND pull for a client, then I can see that but I still don't see how that would happen in parallel.

morango/sync/syncsession.py
```diff
@@ -376,42 +384,139 @@ def _pull_record_chunk(self, chunk_size, transfer_session):
...
+        """
+        self.completed.fire()
+
+
 class SyncClient(object):
     """
     Controller to support client in initiating syncing and performing related operations.
     """
 
+    session = SyncSignalGroup()
```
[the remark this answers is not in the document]
bjester: Good idea

morango/sync/syncsession.py
```diff
@@ -376,42 +384,139 @@ def _pull_record_chunk(self, chunk_size, transfer_session):
...
+    """Signal group firing for each push and pull `TranferSession`."""
+    queuing = SyncSignalGroup(local=True)
```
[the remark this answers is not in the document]
bjester: Ah yes, this would be better to have `local=None`. I think I'll just change them to use `transfer_session`.

morango/proquint.py
```diff
@@ -4,6 +4,7 @@
 functions. For tighter control over the output, see :class:`HumanHasher`.
 """
 import uuid
+import six
```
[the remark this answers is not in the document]
bjester: Removed it.

morango/sync/syncsession.py
```diff
@@ -376,42 +384,139 @@ def _pull_record_chunk(self, chunk_size, transfer_session):
...
+        Fires the `started` signal.
+        """
+        self.started.fire()
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        """
+        Fires the `completed` signal.
+        """
+        self.completed.fire()
```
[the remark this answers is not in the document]
bjester: At the moment, I have no requirements on how to handle or report an error case. We could avoid firing `complete` this if there's an error?

morango/sync/syncsession.py
bjester: I rearranged the code a little bit and will make sure we're report this information specifically in Kolibri.

morango/sync/syncsession.py
```diff
@@ -376,42 +384,139 @@ def _pull_record_chunk(self, chunk_size, transfer_session):
         return self.session.get(self.urlresolve(api_urls.BUFFER), params=params)
 
 
+class SyncSignal(object):
```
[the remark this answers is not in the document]
bjester: Yeah, this is very similar to Django's. In my experience, I generally try to avoid such things since the reason for dependency on Django isn't specifically for Signals. I think it helps keep this (standalone) package uncoupled and easier to maintain. If you feel strongly, I could use it. I think it would only save a few lines.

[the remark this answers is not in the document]
jamalex: Nope, this sounds good -- Django wouldn't handle the "grouping" stuff itself anyway.

morango/sync/syncsession.py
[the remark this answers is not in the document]
jamalex: I think still firing `completed` would be best, and then we should have the Kolibri code check what the status is when it's updating what it shows. I guess it may also be handled in parallel by the job erroring out.

morango/sync/syncsession.py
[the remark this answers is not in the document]
jamalex: TIL PEP 224

morango/sync/syncsession.py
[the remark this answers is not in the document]
jamalex: It probably won't be possible via Kolibri's UI, given how we process the job queue serially, but we can also initiate this syncing via management commands, and could initiate many in parallel. But since each of those is starting up in a separate process, I assume these signals would be contextualized to the appropriate task.

setup.py
```diff
@@ -32,7 +32,6 @@
         "django-ipware>=1.1.6,<1.2",
         "future==0.16.0",
         "requests",
-        "six",
```
jamalex: I think we may then need to change the import paths in the code to import from Django's vendored version, right?

bjester: I grepped and I think I got the only one. It's possible I missed one.

bjester: I mistakenly didn't get that in the same commit though

bjester: Ah looks like I did miss two imports :man_facepalming: Fixing!

bjester:
@jamalex Okay, this is now ready for merge! Let me know what I can do to help release a new version.

Issue #79, opened 2020-02-24
Instance ID changes too frequently


jamalex:
We currently calculate the instance_id (InstanceIDModel) with the utmost priority placed on it being _as unique as possible_. This is because having two devices with the same instance ID generating data in parallel will cause problems if their respective data is ever synced to the same place. 

We currently do this by incorporating as many sources of potential "device uniqueness" as possible into a big hash, in hopes that no two devices will have the same combination of all these values:
https://github.com/learningequality/morango/blob/5da47d241aeb7c48deed681a54589f62bdd28ede/morango/models/core.py#L118

This has served Morango's purposes well. The instance_id can change on moderately frequent intervals (e.g. if Python or kernel is upgraded), but that has no ill effects other than very nominal increases in database size. However, we've also started leaning on the instance_id for other purposes that would do better with more instance_id stability over time, including:
- Nutritionfacts installation counts
- Peer device identification on local network

jamalex:
For this reason, we're proposing updating the calculation of instance_id to use the following flow:

First, calculate the instance_id based on the old method. If it matches the current instance_id, keep using that. This will avoid unnecessary churn if the id wasn't going to change anyway.

Otherwise, to calculate a "new" instance_id:

A. Calculate a system identifier, in the following order of preference:
  - The `MORANGO_SYSTEM_ID` environment variable.
  - The ID that is returned from an OS-specific serial number/UUID system call 
    - Android: `getprop ro.serialno` or `getprop ril.serialnumber` or `getprop ro.boot.serialno` or some [Java code](https://stackoverflow.com/a/5626208)
    - Windows: `wmic csproduct get UUID` (discard if FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF, possibly falling back to `wmic DISKDRIVE get SerialNumber`)
    - Linux: Read `/etc/machine-id`, but that could be duplicated across disk image clones. May be best to just skip this one.
    - MacOS: `ioreg -l | grep IOPlatformSerialNumber` or `subprocess.check_output("ioreg -rd1 -c IOPlatformExpertDevice | grep -E '(UUID)'", shell=True).split('"')[-2]`
  - The MAC address(es) of network adapters. A generic proxy for this is `uuid.getnode()`, as long as the bit is set to show it's from a "real" MAC. May want to use `"-".join(sorted([c["ether"] for c in ifcfg.interfaces().values() if c.get("ether")]))` instead, so it's more robust (uses multiple adapters). Hash this.

[...]

PR #83, opened 2020-06-09
Instance id stability

jamalex:
## Summary

Addresses https://github.com/learningequality/morango/issues/79. Instance ID is now calculated as a function purely of the System ID, MAC address, and Device ID. If it wouldn't otherwise have changed, it is left the same, however.

## TODO

## Reviewer guidance

## Issues addressed

## Documentation


tests/testapp/tests/test_controller_methods.py
```diff
@@ -7,6 +7,7 @@
 from facility_profile.models import Facility
```
jamalex, line 7: Can safely ignore this file. It's primarily just black/linting stuff, and a few small test fixes.

morango/models/core.py
```diff
@@ -115,49 +113,55 @@ class InstanceIDModel(UUIDModelMixin):
...
+    @classmethod
     @transaction.atomic
-    def get_current_instance_and_increment_counter():
-        InstanceIDModel.objects.filter(current=True).update(counter=F("counter") + 1)
-        return InstanceIDModel.objects.get(current=True)
+    def get_current_instance_and_increment_counter(cls):
+        instance, _ = cls.get_or_create_current_instance()
+        cls.objects.filter(id=instance.id).update(counter=F("counter") + 1)
+        instance.refresh_from_db(fields=["counter"])
+        return instance
```
jamalex, line 164: These are the changes that should fix the ["more than one InstanceIDModel" issue](https://github.com/learningequality/morango/issues/79#issuecomment-628177551).

morango/models/core.py
```diff
@@ -115,49 +113,55 @@ class InstanceIDModel(UUIDModelMixin):
     system_id = models.CharField(max_length=100, blank=True)
 
     @classmethod
+    @transaction.atomic
```
jamalex, line 116: The changes in the `get_or_create_current_instance` method, below, implement the behavior described in  https://github.com/learningequality/morango/issues/79 (leveraging the helper functions in `utils.py`)

First, it sees whether the current system parameters (calculated using the "old method") match the "current" instance. If they do, we stick with that to minimize unnecessary churn. If not, we use the "new method" to compute a more robust instance ID.

morango/models/utils.py
```diff
@@ -0,0 +1,197 @@
+import hashlib
```
jamalex, line 1: This file is where a lot of the real "work" happens, to support calculation of the Instance ID.

tests/testapp/tests/helpers.py
```diff
@@ -11,6 +11,7 @@
 from facility_profile.models import InteractionLog
```
jamalex, line 11: I had to update a number of the tests to change the mocking, since we now compute the Instance ID as a function of different things. This file is an example of that (most of the change is from indentation).

tests/testapp/tests/test_uuid_utilities.py
```diff
@@ -1,29 +1,43 @@
+import hashlib
```
jamalex, line 1: Many of the changes here are just from running "black". I'll flag the new stuff.

tests/testapp/tests/test_uuid_utilities.py
```diff
@@ -52,39 +65,109 @@ def test_creating_same_instance_ID_model(self):
...
+        with mock.patch(
+            "uuid.getnode", return_value=67002173923623
+        ):  # fake (random) address
             (IDModel, _) = InstanceIDModel.get_or_create_current_instance()
 
-        self.assertFalse(InstanceIDModel.objects.exclude(id=ident).filter(current=True).exists())
+        self.assertFalse(
+            InstanceIDModel.objects.exclude(id=ident).filter(current=True).exists()
+        )
         self.assertTrue(InstanceIDModel.objects.get(id=ident).current)
```
jamalex, line 94: The two tests below are new, and are designed to ensure that the Instance ID doesn't change when we aren't wanting it to.

morango/models/core.py
```diff
@@ -115,49 +113,55 @@ class InstanceIDModel(UUIDModelMixin):
...
+        )
+        # create the model, or get existing if one already exists with this ID
+        obj, created = InstanceIDModel.objects.get_or_create(
+            id=kwargs["id"], defaults=kwargs
+        )
+        # update any of the attributes that have changed
+        changed = False
+        for key, val in kwargs.items():
+            if getattr(obj, key) != val:
+                setattr(obj, key, val)
```
bjester, line 151: Would using `update_or_create` instead of `get_or_create` work the same?

morango/models/utils.py
```diff
@@ -0,0 +1,197 @@
...
+    Sort interfaces by device name to give preference to interfaces
+    that are more likely to be stable (not change/be removed).
+    """
+    dev = (iface.get("device") or "").lower()
+    if dev.startswith("eth") or dev.startswith("en"):
+        return "0" + dev
+    if dev.startswith("wl"):
+        return "1" + dev
+    if dev.startswith("e") or dev.startswith("w"):
+        return "2" + dev
```
bjester, line 173: Would it be helpful to exclude bridged connections entirely `dev.startswith("br")`? I know with specific Docker network configurations Kolibri would fail to start because of something to do with the mac addresses.

morango/models/core.py
jamalex: Yes, I think so! Had forgotten about that one.

morango/models/utils.py
jamalex: I also noticed a vbox*** adapter that had a MAC address, while testing (as I had a virtualbox running in parallel). Don't want to include those either. Wasn't sure I could come up with an exhaustive list of "non-real" interface names, though... especially in a cross-platform way. Any ideas?

jamalex: I'm adding something to check flags in the MAC address for whether it's "local", vs a registered physical address, based on this:
https://en.wikipedia.org/wiki/MAC_address#Universal_vs._local

That seems to exclude at least the vbox case, and should prevent having it pick up on a transient, generated MAC, which would be bad for ID stability.

bjester: Cool. The exact issue I recall is when you run a docker container with host networking mode, even if it wasn't running, it would cause issues because it was actually the same MAC address as the real adapter. I don't think that should necessarily change what you're doing though.

jamalex: Urgh, yeah, looks as if this was never resolved upstream in ifcfg:
https://github.com/ftao/python-ifcfg/issues/34

The "fix" in Kolibri, apparently by me, was to just... ignore it:
https://github.com/learningequality/kolibri/pull/5914/files

:sigh:

jamalex: I was unable to replicate that issue, and we at least handle it gracefully here, so I don't think it's a blocker at the moment.

PR #84, opened 2020-06-11
Deserialization improvements

jamalex:
## Summary

Some records at times fail to deserialize on KDP (due to missing data, specifically foreign key referents). Added some tests that uncovered a potential cause, and fixed that, and then also added a column called `deserialization_error` on the `Store` model to track issues with deserialization, to support debugging and to be able to skip them for a "lighter weight" deserialization as desired.

## TODO

## Issues addressed

Addresses issues with bulk deserialization seen on KDP.


jamalex:
> The only thing that confused me initially was that `_deserialize_store_model` catches and _re-raises_ the error.

Yeah, not immediately obvious from looking at the code what's going on there. I'll add a couple more comments to that section.

PR #59, opened 2018-06-27
Handle network exceptions

ralphiee22:
## Summary

First pass of retry/resume logic on syncing sessions for flaky connections/power dropouts/other errors.

## TODO

- [x] Have tests been written for the new code?
- [ ] Has documentation been written/updated?
- [ ] New dependencies (if any) added to requirements file

## Reviewer guidance

*If you PR has a significant size, give the reviewer some helpful remarks*

## Issues addressed

List the issues solved or partly solved by the PR

## Documentation

If the PR has documentation, link the file here (either .rst in your repo or if built on Read The Docs)

PR #86, opened 2020-07-16
[WIP] Enable resuming of sync sessions

bjester:
## Summary

*Short description*

## TODO

## Reviewer guidance

## Issues addressed

## Documentation


morango/api/viewsets.py
```diff
@@ -275,19 +271,44 @@ def create(self, request):
             "connection_path": request.data.get("connection_path"),
             "client_ip": get_ip(request) or "",
             "server_ip": request.data.get("server_ip") or "",
-            "client_instance": request.data.get("instance"),
-            "server_instance": json.dumps(
-                serializers.InstanceIDSerializer(instance_id).data
-            ),
```
jamalex, line 281: In terms of resuming sessions, we actually care more about the client/server instance_ids being the same, than we do about the IPs being the same (better indicator that the same session is likely to also exist on the other side). Since here we have the full dict (not just the instance_id), it would be hard to filter by efficiently, though. And the IPs may change, but not super frequently, so filtering by IP is probably a good middle ground. We just need to make sure we're smoothly handling cases where the session doesn't actually exist on the other side, since IPs may be reused.

morango/api/viewsets.py
```diff
@@ -275,19 +271,44 @@ def create(self, request):
...
-        syncsession.full_clean()
-        syncsession.save()
+        sync_session = None
+
+        # if we're allowing resume, then try to pull an existing session
+        if allow_resume:
+            query = (
+                core.SyncSession.objects.filter(**data)
+                .order_by("last_activity_timestamp")
+                .reverse()
```
jamalex, line 284: Can also do `.order_by("-last_activity_timestamp")`

jamalex:
I hadn't realized that we never actually merged this old PR, which contained a lot of network error handling and session resumption code: https://github.com/learningequality/morango/pull/59

Haven't had a chance to fully review here, and I know things have already diverged in mainline since #59 was started, but it may be worth us reviewing to see if there are pieces we can pull in from there or some way to integrate (as a lot of work and iteration went into that one). Not urgent, though -- maybe let's revisit after 0.14 is out?

PR #87, opened 2020-07-17
Cache the instance ID on app load, to avoid lockup issues we were seeing

jamalex:
## Summary

To hopefully address https://github.com/learningequality/kolibri/issues/7284

## TODO

### DIFF
--- a/morango/models/core.py
+++ b/morango/models/core.py
@@ -91,6 +91,8 @@ class InstanceIDModel(models.Model):
     as well as its counter with all the records that were serialized at the time.
     """
 
+    _cached_instance = None
+
     uuid_input_fields = (
         "platform",
         "hostname",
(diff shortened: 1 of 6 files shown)


morango/models/core.py
```diff
@@ -114,37 +116,57 @@ class InstanceIDModel(models.Model):
...
+                instance.refresh_from_db(fields=["counter", "current"])
+                # only use cached instance if it's still marked as current, otherwise skip
+                if instance.current:
+                    return cls._cached_instance, False
+            except InstanceIDModel.DoesNotExist:
+                # instance does not exist, so skip here so we create a new one
+                pass
+
+        with transaction.atomic():
+
```
jamalex, line 139: Most of the stuff below hasn't changed, except:
- Indenting to put it into the transaction block
- Renaming "obj" to "instance" for clarity
- Setting the cache value before returning

morango/models/core.py
```diff
@@ -114,37 +116,57 @@ class InstanceIDModel(models.Model):
...
-        obj, created = InstanceIDModel.objects.update_or_create(
-            id=kwargs["id"], defaults=kwargs
-        )
-        return obj, created
+        if cls._cached_instance:
+            instance = cls._cached_instance
+            # make sure we have the latest counter value and "current" flag
+            try:
+                instance.refresh_from_db(fields=["counter", "current"])
+                # only use cached instance if it's still marked as current, otherwise skip
```
jamalex, line 131: In general this shouldn't happen, but it did turn up in a test (because of how tests reset the database without clearing caches), and this will make it more robust to edge cases.

morango/models/core.py
```diff
@@ -114,37 +116,57 @@ class InstanceIDModel(models.Model):
...
+                if instance.current:
+                    return cls._cached_instance, False
+            except InstanceIDModel.DoesNotExist:
+                # instance does not exist, so skip here so we create a new one
+                pass
```
jamalex, line 136: In general this exception shouldn't happen, but it did turn up in a test (because of how tests reset the database without clearing caches), and this will make it more robust to edge cases.

jamalex:
@rtibbles Another thing to note: I used a class-level in-memory cache, because it avoids needing to pickle or whatnot to store it in a Django cache (we want the model, not just the instance_id). Also, the uwsgi case, if we're spinning up multiple processes, would have cleared the cache on app load anyway, so it wouldn't have increased efficiency.

jamalex:
Pushed a change based on discussion in #morango -- hoping it'll pass now

PR #90, opened 2020-07-31
Allow MAC address to be overridden by envvar for instance_id calculation

jamalex:
## Summary

On KDP (or in Docker-based distributions more broadly), the detected MAC address is randomly generated, leading to regeneration of the instance_id more frequently than is desired (and in some cases, database lockups, in a multi-node setup). This PR allows the detected MAC value to be overridden with a fixed value, to increase the instance_id stability in these environments.

## TODO

### DIFF
--- a/morango/models/utils.py
+++ b/morango/models/utils.py
@@ -205,6 +205,11 @@ def _mac_is_local(mac):
 
 def get_0_5_mac_address():
 
+    # check whether envvar was set, and use that if available
+    node_id = os.environ.get("MORANGO_NODE_ID")
+    if node_id and len(node_id.strip()) >= 3:
+        return _do_salted_hash(node_id)
+
     # first, try using ifcfg
     interfaces = []
     try:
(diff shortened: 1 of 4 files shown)


jamalex:
This has been released, so it can be used on KDP, but leaving open here for review.

morango/models/utils.py
```diff
@@ -205,6 +205,11 @@ def _mac_is_local(mac):
 
 def get_0_5_mac_address():
 
+    # check whether envvar was set, and use that if available
+    node_id = os.environ.get("MORANGO_NODE_ID")
+    if node_id and len(node_id.strip()) >= 3:
+        return _do_salted_hash(node_id)
```
indirectlylit, line 211: cool, glad we're salting and hashing it

indirectlylit: > `len(node_id.strip()) >= 3`

maybe we should provide a warning the case that `MORANGO_NODE_ID` exists but is being ignored

jamalex: Opened an issue: https://github.com/learningequality/morango/issues/91

PR #96, opened 2020-10-30
docs updates

indirectlylit:
minor updates to docs made as I try to understand the system on a deeper level


indirectlylit:
migrating work to #97 and continuing there

PR #112, opened 2021-05-06
Set up GH workflows

bjester:
## Summary
Replaces Travis with Github workflows. I had to remove testing for all but one existing version of `cryptography` as it seemed those versions were just too old. I also resolved issues with Python 3.9

[build](https://github.com/bjester/morango/actions) 
[build](https://github.com/bjester/morango/actions)
[build](https://github.com/bjester/morango/actions)

## Reviewer guidance
Python unit test workflow result: https://github.com/bjester/morango/actions/runs/818473128
Python linting workflow result: https://github.com/bjester/morango/actions/runs/818473130
Build docs workflow result: https://github.com/bjester/morango/actions/runs/818473129


rtibbles (approved):
This all looks good to me - we may want to add Windows and MacOS tests in the future as well, but let's not let the perfect be the enemy of the good!

PR #120, opened 2021-06-25
Add management command to close old sessions and clear out the buffers

jamalex:
## Summary

If sync sessions are interrupted, or time out, they may not be closed out successfully, leaving possibly large amounts of data in the buffers. This management command can be called to close out and clean up after sync sessions that are older than a certain number of hours.

## TODO

### DIFF
--- a/morango/management/commands/cleanupsyncs.py
+++ b/morango/management/commands/cleanupsyncs.py
@@ -1,0 +1,60 @@
+import datetime
+
+from django.core.management.base import BaseCommand
+from django.db import connection
+from django.db import transaction
+from django.utils import timezone
+
+from morango.models import *
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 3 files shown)


morango/management/commands/cleanupsyncs.py
```diff
@@ -0,0 +1,60 @@
...
+            with transaction.atomic():
+                with connection.cursor() as cursor:
+                    cursor.execute(
+                        "DELETE FROM morango_buffer WHERE transfer_session_id = %s",
+                        (sess.id,),
+                    )
+                    cursor.execute(
+                        "DELETE FROM morango_recordmaxcounterbuffer WHERE transfer_session_id = %s",
+                        (sess.id,),
+                    )
```
jamalex, line 56: Doing the raw SQL like this was about 15 times faster than doing bulk deletion via the ORM.

bjester (approved):
Looks good to me

PR #122, opened 2021-06-26
Tag records in the Store with the last transfersession that modified them

jamalex:
## Summary

In Kolibri 0.15, we'll need to do some "post-sync" operations (e.g. for generating notifications) that can be optimized by knowing exactly which records were just synced in. This PR adds logic to tag records in the `Store` with the ID of the last transfer session that touched them. If they're overwritten from the app models by serializing, this value is reset to `None`.

A utility function on `TransferSession` called `get_touched_record_ids_for_model` returns a list (values queryset, actually, to aid a direct join) of record IDs that were modified by that transfer session. Currently it doesn't distinguish between whether the record was updated, inserted, or deleted, but the first two can be distinguished from the third by querying the Store to see if it's flagged as deleted, or by the absence of the record in the app models.

## TODO


bjester (approved):
Looks good to me!

Issue #119, opened 2021-06-25
`last_activity_timestamp` values aren't updated on session activity


jamalex:
We set this field on the `SyncSession` and `TransferSession` when they're first created ([here](https://github.com/learningequality/morango/blob/380cab228a72a0ac6a20926ae6963cb76054b9e1/morango/sync/syncsession.py#L197) and [here](https://github.com/learningequality/morango/blob/380cab228a72a0ac6a20926ae6963cb76054b9e1/morango/sync/syncsession.py#L545)), but I can't find anywhere that they get updated as a side-effect of ongoing activity within the sessions.

This is corroborated by this KDP data that shows the start and last activity timestamps being near identical across the board:
```
>>> while True: ss = SyncSession.objects.order_by("?")[0]; ts = ss.transfersession_set.first(); ts.last_activity_timestamp - ts.start_timestamp
datetime.timedelta(microseconds=11)
datetime.timedelta(microseconds=54)
datetime.timedelta(microseconds=8)
datetime.timedelta(microseconds=6)
datetime.timedelta(microseconds=29)
datetime.timedelta(microseconds=8)
datetime.timedelta(microseconds=8)
datetime.timedelta(microseconds=8)
datetime.timedelta(microseconds=8)
datetime.timedelta(microseconds=9)
datetime.timedelta(microseconds=9)
datetime.timedelta(microseconds=6)
datetime.timedelta(microseconds=11)
datetime.timedelta(microseconds=9)
datetime.timedelta(microseconds=8)
datetime.timedelta(microseconds=6)
datetime.timedelta(microseconds=9)
datetime.timedelta(microseconds=7)
datetime.timedelta(microseconds=9)
datetime.timedelta(microseconds=14)
[...]
```

This field isn't currently being actively referenced anywhere in code, so it's not a critical issue at the moment, but this will become more important once the garbage collection code comes in (soon), so I wanted to make sure we tracked this.

bjester:
Resolved by https://github.com/learningequality/morango/pull/125

PR #125, opened 2021-07-15
Trigger update to `last_activity_timestamp`

bjester:
## Summary
Ensures `last_activity_timestamp` on both `SyncSession` and `TransferSession` are updated when "activity" occurs. In this case, activity is registered when a `TransferSession` progresses through transfer stages.

## TODO

## Reviewer guidance
Does `last_activity_timestamp` get updated as a sync progresses?

## Issues addressed
Fixes https://github.com/learningequality/morango/issues/119


rtibbles (approved):
Code makes sense, as do the assertions the tests are making!

PR #128, opened 2021-07-22
Morango 0.6.1 fixes

bjester:
## Summary
- Don't boolean compare against `get_filter()` as it always returns a `Filter` object
- Raise error in viewset cleanup if set on context for better insights
- Debug logging to track invocation and results of operations
- Set counters on `TransferSession` after serialization
- Fix capabilities header as it should be prefixed with `HTTP`
- Fix issues with flow of operations

## TODO

## Reviewer guidance
From Kolibri, testing with `INTEGRATION_TEST=1 pytest -x kolibri/core/auth/test/test_morango_integration.py ` was helpful

### DIFF
--- a/morango/sync/context.py
+++ b/morango/sync/context.py
@@ -52,7 +52,7 @@ class SessionContext(object):
         if self.transfer_session:
             self.sync_session = transfer_session.sync_session or self.sync_session
             self.is_push = transfer_session.push or self.is_push
-            if transfer_session.filter and transfer_session.filter is not None:
+            if transfer_session.filter:
                 self.filter = transfer_session.get_filter()
 
     def update(
@@ -101,7 +101,7 @@ class SessionContext(object):
         if transfer_session:
             self.sync_session = transfer_session.sync_session
             self.is_push = transfer_session.push
-            if transfer_session.filter and transfer_session.filter is not None:
+            if transfer_session.filter:
                 self.filter = transfer_session.get_filter()
 
     @property
--- a/morango/sync/context.py
+++ b/morango/sync/context.py
@@ -51,8 +51,9 @@ class SessionContext(object):
 
         if self.transfer_session:
             self.sync_session = transfer_session.sync_session or self.sync_session
-            self.filter = transfer_session.get_filter() or self.filter
             self.is_push = transfer_session.push or self.is_push
+            if transfer_session.filter and transfer_session.filter is not None:
+                self.filter = transfer_session.get_filter()
 
     def update(
         self,
@@ -99,8 +100,9 @@ class SessionContext(object):
         # if transfer session was passed in, that takes precedence
         if transfer_session:
             self.sync_session = transfer_session.sync_session
-            self.filter = transfer_session.get_filter() or self.filter
             self.is_push = transfer_session.push
+            if transfer_session.filter and transfer_session.filter is not None:
+                self.filter = transfer_session.get_filter()
 
     @property
     def is_pull(self):
--- a/morango/sync/controller.py
+++ b/morango/sync/controller.py
@@ -242,7 +242,7 @@ class SessionController(object):
             return result
         except Exception as e:
             # always log the error itself
-            logging.error(e)
+            logger.error(e)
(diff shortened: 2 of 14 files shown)


morango/sync/controller.py
```diff
@@ -242,7 +242,7 @@ def _invoke_middleware(self, context, middleware):
             return result
         except Exception as e:
             # always log the error itself
-            logging.error(e)
+            logger.error(e)
```
rtibbles, line 245: If I had a nickel...

morango/sync/context.py
```diff
@@ -51,8 +51,9 @@ def __init__(
 
         if self.transfer_session:
             self.sync_session = transfer_session.sync_session or self.sync_session
-            self.filter = transfer_session.get_filter() or self.filter
             self.is_push = transfer_session.push or self.is_push
+            if transfer_session.filter and transfer_session.filter is not None:
```
jamalex, line 55: If it's `None`, wouldn't it have already failed the falsiness check in the first half of the `and`? So the second half may not be necessary?

morango/sync/syncsession.py
```diff
@@ -690,17 +698,6 @@ def __init__(self, *args, **kwargs):
         self.local_context.update(is_push=True)
         self.remote_context.update(is_push=True)
 
-    def initialize(self, sync_filter):
-        """
-        :param sync_filter: Filter
-        """
-        super(PushClient, self).initialize(sync_filter)
-
-        self.sync_connection._update_transfer_session(
```
rtibbles, line 699: Where does the records_total get set now?

rtibbles (commented):
Just one question about the records_total saving.

morango/sync/syncsession.py
bjester: It's in the [`NetworkQueueOperation`](https://github.com/learningequality/morango/pull/128/files#diff-dce2c6c78fa132287ff3004fb1a3eed42377dc787b8bf4986d544d501be22c9dR1121) which should occur after queuing locally

morango/sync/syncsession.py
```diff
@@ -634,8 +639,11 @@ def initialize(self, sync_filter):
             transfer_session=self.current_transfer_session
         )
 
-        # create transfer session on server side and proceed to queuing
+        # backwards compatibility for the queuing signal as it included both serialization
```
jamalex, line 642: Since Kolibri is the only (known) consumer of Morango, we'd probably be fine just updating it on the Kolibri side, too, and not worrying about local-only API changes (which I think these signals may be); just need to worry about things that would impact one version of Kolibri talking to another.

(fine to leave like this, though -- but thought it worth noting in case it simplifies)

morango/sync/utils.py
```diff
@@ -104,30 +102,24 @@ def validate_and_create_buffer_data(  # noqa: C901
...
         if not transfer_session.get_filter().contains_partition(record["partition"]):
             raise ValidationError(
-                {
-                    "partition": "Partition {} is not contained within filter for TransferSession ({})".format(
-                        record["partition"], transfer_session.filter
-                    )
-                }
+                "Partition {} is not contained within filter for TransferSession ({})".format(
+                    record["partition"], transfer_session.filter
+                )
```
jamalex, line 107: This didn't click for me when you were showing the Exception format the other day, but I remember now that this method of passing a dict to a `ValidationError` is/was a Django thing, where it tells the frontend which field/attribute of the input data the error is associated with (e.g. for rendering forms, so it knows which input box to position the red error message next to). Is that no longer the case?

(Mostly a curiosity, since these aren't models that we'd be editing manually via forms anyway, so this change is good.)

jamalex (approved):
Looks good to me. Added a couple small notes/questions, but not blockers.

morango/sync/utils.py
bjester: Since `validate_and_create_buffer_data` is now only ever called in an operation, the dictionary format no longer serves a purpose.

jamalex: Makes sense -- and I guess ValidationError still makes sense, since what it's doing is... validating :)

PR #129, opened 2021-07-26
Fix handling of sleep for `proceed_to_and_wait_for`

bjester:
## Summary
I forgot to circle back around to ensuring the sleeping interval between checks on an operation's status is optimal. The result was that we were sleeping a lot, too much, and it slowed down syncing quite a bit.

### Profiling screenshots
| Before | After |
| --- | ---
| Screenshot from 2021-07-26 16-14-01 | Screenshot from 2021-07-26 16-44-51 |

## TODO

## Reviewer guidance
You probably can't tell from the tests that it will be faster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 List of the most important changes for each release.
 
+## 0.6.2
+
+- Fixes slow performance due to excessive use of `sleep`
+
 ## 0.6.1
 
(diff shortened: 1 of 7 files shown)


rtibbles (approved):

jamalex (approved):
Looks great, and makes sense! We might find it useful later to make these params configurable (fallback, maxes, etc), but these seem like sane defaults.

PR #131, opened 2021-08-13
Replace assert statements to avoid removal with python -O

bjester:
## Summary

- Python 4 android compiles `.py` files into `.pyo` by default which removes plain `assert` statements, the same behavior as `-O` flag
- Some other tweaks from Kolibri integration

## TODO

## Reviewer guidance
- Are you able to sync Kolibri using `python -O`

### DIFF
--- a/morango/api/permissions.py
+++ b/morango/api/permissions.py
@@ -33,7 +33,7 @@ class BasicMultiArgumentAuthentication(authentication.BasicAuthentication):
                 credentials[key] = val
 
         # authenticate the user via Django's auth backends
-        user = authenticate(**credentials)
+        user = authenticate(request=request, **credentials)
 
         if user is None:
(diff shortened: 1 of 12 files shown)


morango/api/permissions.py
```diff
@@ -33,7 +33,7 @@ def authenticate_credentials(self, userargs, password, request=None):
                 credentials[key] = val
 
         # authenticate the user via Django's auth backends
-        user = authenticate(**credentials)
+        user = authenticate(request=request, **credentials)
```
bjester, line 36: Seemed like a good idea to make sure this passes along the request object, but it isn't required.

rtibbles: Seems sensible, and it's not in the credentials dict, so seems low risk!

rtibbles (approved):

PR #132, opened 2021-08-18
Chunk through fsics when queuing into the buffer

bjester:
## Summary
- Fixes `Expression tree is too large (maximum depth 1000)` issue originating from client or server having many instance counters

## TODO

## Reviewer guidance
- New test fails without change

## Issues addressed

Resolves https://github.com/learningequality/morango/issues/114

### DIFF
--- a/morango/sync/operations.py
+++ b/morango/sync/operations.py
@@ -413,7 +413,6 @@ def _queue_into_buffer(transfersession):
     We use raw sql queries to place data in the buffer and the record max counter buffer, which matches the conditions of the FSIC,
     as well as the partition for the data we are syncing.
     """
-    last_saved_by_conditions = []
     filter_prefixes = Filter(transfersession.filter)
     server_fsic = json.loads(transfersession.server_fsic)
     client_fsic = json.loads(transfersession.client_fsic)
@@ -427,18 +426,6 @@ def _queue_into_buffer(transfersession):
     if not fsics:
         return
 
-    # create condition for all push FSICs where instance_ids are equal, but internal counters are higher than FSICs counters
-    for instance, counter in six.iteritems(fsics):
-        last_saved_by_conditions += [
-            "(last_saved_instance = '{0}' AND last_saved_counter > {1})".format(
-                instance, counter
-            )
-        ]
-    if fsics:
-        last_saved_by_conditions = [
-            _join_with_logical_operator(last_saved_by_conditions, "OR")
-        ]
-
     partition_conditions = []
     # create condition for filtering by partitions
     for prefix in filter_prefixes:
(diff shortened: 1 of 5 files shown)


rtibbles (approved):

jamalex:
The overall approach looks in line with what I'd been imagining as well. The only thing that looking through it raised for me is something I hadn't considered before: the separate querying and queuing by chunks of instance IDs has the potential to queue the same records multiple times within the same transfer session. Specifically, imagine this scenario:

- Device A syncs Record X to both Device B and Device C.
- Device B modifies the record, then syncs to C.
- Device C further modifies the record, then syncs back to A.
- During this last sync, imagine that the instance IDs for Device B and C are in different chunks of the FSIC.
- The `last_saved_by_conditions` for the syncing of both chunks will match Record X, and so it will be queued twice (once during each chunked query).

I don't think this should cause any catastrophic failures, but it definitely seems like something we'd want to avoid (data bloat, etc, or maybe other unforeseen consequences resulting from the expectation that this wouldn't happen). You may have other ideas for addressing it, but it may be enough to add an additional condition to the query along the lines of (warning: pseudocode) `AND store.id NOT IN (SELECT model_uuid from buffer where buffer.transfer_session_id = '{transfer_session_id}')`.

Happy to co-hack as helpful!

jamalex (changes requested):
The overall approach looks in line with what I'd been imagining as well. The only thing that looking through it raised for me is something I hadn't considered before: the separate querying and queuing by chunks of instance IDs has the potential to queue the same records multiple times within the same transfer session. Specifically, imagine this scenario:

- Device A syncs Record X to both Device B and Device C.
- Device B modifies the record, then syncs to C.
- Device C further modifies the record, then syncs back to A.
- During this last sync, imagine that the instance IDs for Device B and C are in different chunks of the FSIC.
- The `last_saved_by_conditions` for the syncing of both chunks will match Record X, and so it will be queued twice (once during each chunked query).

I don't think this should cause any catastrophic failures, but it definitely seems like something we'd want to avoid (data bloat, etc, or maybe other unforeseen consequences resulting from the expectation that this wouldn't happen). You may have other ideas for addressing it, but it may be enough to add an additional condition to the query along the lines of (warning: pseudocode) `AND store.id NOT IN (SELECT model_uuid from buffer where buffer.transfer_session_id = '{transfer_session_id}')`.

Happy to co-hack as helpful!

morango/sync/operations.py
```diff
@@ -427,66 +426,101 @@ def _queue_into_buffer(transfersession):
...
-            "profile = '{}'".format(transfersession.sync_session.profile),
-        ],
-        "AND",
+    # chunk fsics creating multiple SQL selects which will be unioned before insert
+    i = 0
+    chunk_size = 200
+    chunk = fsics[:chunk_size]
+    select_buffers = []
+    select_rmc_buffers = []
+    transfersession_pk_field = next(
```
rtibbles, line 445: Can also just use `tranfersession_pk_field = TransferSession._meta.pk` here.

tests/testapp/tests/sync/test_operations.py
```diff
@@ -86,6 +86,22 @@ def test_all_fsics(self):
...
         self.assertRecordsBuffered(self.data["group2_c1"])
 
+    def test_very_many_fsics(self):
+        """
+        Regression test against 'Expression tree is too large (maximum depth 1000)' error with many fsics
+        """
+        fsics = {self.data["group1_id"].id: 1, self.data["group2_id"].id: 1}
+        fsics.update({
+            uuid.uuid4().hex: i
+            for i in range(2000)
```
jamalex, line 96: Since we're packing it all back into a single query now, I wonder if it might be good to increase the size of this by an order of magnitude just to be sure we're not just kicking the can down the road?

jamalex: (Particularly as there may be maximum query lengths in SQLite)

jamalex (changes requested):
The approach looks sound to me if it does in practice overcome the size limits effectively -- I added one comment in there around increasing the size of the test data to push the limits a bit further.

I'm still not certain (not knowing UNION well) whether this will deduplicate for the conditions I described previously (as each chunk's SELECT is separate and then joined, not sure if it does DISTINCT on that), so adding a regression test could be good, just to be sure. If there's not clean/efficient way to dedupe this approach, I also suggested a possible alternative based on the previous chunked query approach.

jamalex (approved):
Looks great! Thanks!

PR #134, opened 2021-09-20
Add asymmetry to FSIC calculation to ensure all matching data is synced

jamalex:
## Summary

Function `calculate_filter_specific_instance_counters` returns a dict that maps instance_ids
to their respective "high-water level" counters with respect to the provided list of filter
partitions, based on what the local database contains.

First, for each partition in the filter, it calculates the maximum values the database has
received through any filters containing that partition.

Then, it combines these dicts into a single dict, collapsing across the filter partitions.
In Morango 0.6.5 and below, this was always calculated based on the "minimum" values for
each instance_id, and with instance_ids that didn't exist in *each* of the partitions being
excluded entirely. When the producing side had records needing to be sent for an instance
under one of the filter partitions, but not under another, it would not be included in the
FSIC and thereby lead to the data not being sent, as showed up in:
https://github.com/learningequality/kolibri/issues/8439

The solution was to add an asymmetry in how FSICs are calculated, with the sending side
using a "max" instead of a "min" to ensure everything is included, and then the receiving
side still using a "min" (though after it has completed a sync, it updates its counters
such that the min and max should then be equivalent).

One potential issue remains, but it is an edge case that can be worked around:
- We now take the maxes across the filter partitions and use those as the producer FSICs.
- When the receiver finishes integrating the received data, it updates its counters to match.
- If the sender had actually done a sync with just a subset of those filters in the past, it
  might not actually have received everything available for the other filters, and hence the
  receiver may not be correct in assuming it now has everything up to the levels of the
  producer's FSICs (as it does by taking the "max" across the filter partition FSICs).

There are two ways to avoid this:
- Don't sync with differing subsets of the same partitions across multiple syncs. For
  example, if you do syncs with filters "AB" and "AC", don't also do syncs with filters
  "AC" and "AD". This is the approach that makes this work in Kolibri, for now.
- OR: Don't do syncs with more than one filter partition at a time. Do each one in sequence.
  For example, rather than pushing "AB" and "AC" in a single transfer session, do one pull
  for AB and then another one for AC. This has the disadvantage of a bit of extra overhead,
  but would likely be the most robust option, and the easiest to enforce and reason about.

## TODO

## Reviewer guidance

The actual meaningful changes are in https://github.com/learningequality/morango/commit/09b17d77607a08653322abd1ef46e85fcb76ba4a -- the other commits are just cleanup/linting and version bumping. A lot of the diff is also just due to the renaming of the function.

## Issues addressed

Towards addressing https://github.com/learningequality/kolibri/issues/8439

### DIFF
--- a/morango/__init__.py
+++ b/morango/__init__.py
@@ -3,4 +3,4 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 default_app_config = "morango.apps.MorangoConfig"
-__version__ = "0.6.5"
+__version__ = "0.6.6"
--- a/morango/__init__.py
+++ b/morango/__init__.py
@@ -3,4 +3,4 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 default_app_config = "morango.apps.MorangoConfig"
-__version__ = "0.6.6-beta"
+__version__ = "0.6.6a0"
--- a/morango/__init__.py
+++ b/morango/__init__.py
@@ -3,4 +3,4 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 default_app_config = "morango.apps.MorangoConfig"
-__version__ = "0.6.6"
+__version__ = "0.6.6-beta"
--- a/morango/models/core.py
+++ b/morango/models/core.py
@@ -11,6 +11,7 @@ from django.db import router
 from django.db import transaction
 from django.db.models import F
 from django.db.models import Func
+from django.db.models import Max
 from django.db.models import TextField
 from django.db.models import Value
 from django.db.models.deletion import Collector
@@ -19,6 +20,7 @@ from django.db.models.functions import Cast
 from django.utils import six
 from django.utils import timezone
 
+from functools import reduce
 from morango import proquint
 from morango.registry import syncable_models
 from morango.models.certificates import Certificate
@@ -527,7 +529,7 @@ class DatabaseMaxCounter(AbstractCounter):
     @classmethod
     @transaction.atomic
     def update_fsics(cls, fsics, sync_filter):
-        internal_fsic = DatabaseMaxCounter.calculate_filter_max_counters(sync_filter)
+        internal_fsic = DatabaseMaxCounter.calculate_filter_specific_instance_counters(sync_filter)
         updated_fsic = {}
         for key, value in six.iteritems(fsics):
             if key in internal_fsic:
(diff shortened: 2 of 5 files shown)


morango/models/core.py
```diff
@@ -546,32 +550,72 @@ def update_fsics(cls, fsics, sync_filter):
...
-        )
+        per_filter_max = []
 
-        with connection.cursor() as cursor:
-            cursor.execute(filter_max_calculation)
-            # try to get hex value because postgres returns values as uuid
-            return {getattr(tup[0], "hex", tup[0]): tup[1] for tup in cursor.fetchall()}
+        for filt in filters:
+            qs = queryset.annotate(filt=Value(filt, output_field=models.CharField()))
+            qs = qs.filter(filt__startswith=F("partition"))
```
bjester: For the above 2 lines, this seems a little bit of Django hand-holding to get the query we want. It doesn't make the intended SQL very clear at first glance, but I don't have a good solution as I've struggled with this before. I would recommend just adding a comment above summarizing the intended resulting SQL.

bjester: Also worth noting Django adds some nested string replacements to the clause which seem unnecessary to me:
```sql
SELECT "morango_databasemaxcounter"."instance_id", MAX("morango_databasemaxcounter"."counter") AS "maxval" 
FROM "morango_databasemaxcounter" 
WHERE %s LIKE REPLACE(REPLACE(REPLACE(("morango_databasemaxcounter"."partition"), \'\\\', \'\\\\\'), \'%%\', \'\\%%\'), \'_\', \'\\_\') || \'%%\' ESCAPE \'\\\' GROUP BY "morango_databasemaxcounter"."instance_id"
```

jamalex: Yeah, it does feel like a case where raw SQL (as used in the original code for this function) might end up being cleaner, I just didn't feel confident getting the result I wanted, given the issues we had with the previous implementation (1: I never understood why `HAVING {count} = COUNT(PMC.filter_partition)` was needed or what exactly it did, and 2: it behaved weirdly, where if you put the same filter in the query twice, it would just return an empty set, with no matches, which wasn't logically consistent with the goal of the query).

My goal here was to get something I knew worked (and that I could wrap my head around sufficiently to reason about), and then use that as a baseline for further optimization. This does make things slower, especially since I'm doing a query per filter, though there are usually only 3 or 4 filters max -- somewhere in the order of 40 microseconds vs 2 milliseconds -- so it would be nice to optimize it again once we've verified this produces the desired outcomes.

What do you think of the idea of me opening a tech debt issue to revisit and clean this up/optimize it?

bjester: It looked like `HAVING {count} = COUNT(PMC.filter_partition)` enforced that all filters matched, since `{count}` was the length of the filters. So essentially what you're doing is `HAVING COUNT(PMC.filter_partition) > 0`, where at least one filter matches.

> so it would be nice to optimize it again once we've verified this produces the desired outcomes.

Sounds like maybe we should make this a beta release then?

This will get us the simplified query that we want:
```diff
diff --git a/morango/models/core.py b/morango/models/core.py
index 520b23c..abfa8e3 100644
--- a/morango/models/core.py
+++ b/morango/models/core.py
@@ -12,11 +12,15 @@ from django.db import transaction
 from django.db.models import F
 from django.db.models import Func
 from django.db.models import Max
+from django.db.models import Q
 from django.db.models import TextField
 from django.db.models import Value
 from django.db.models.deletion import Collector
+from django.db.models.expressions import CombinedExpression
 from django.db.models.fields.related import ForeignKey
 from django.db.models.functions import Cast
+from django.db.models.functions import Concat
+from django.db.models.sql.where import WhereNode
 from django.utils import six
 from django.utils import timezone
 
@@ -514,6 +518,16 @@ class AbstractCounter(models.Model):
         abstract = True
 
 
+class ValueStartsWithField(CombinedExpression):
+    def __init__(self, value, field):
+        super(ValueStartsWithField, self).__init__(
+            Value(value, output_field=models.CharField()),
+            "LIKE",
+            CombinedExpression(F(field), '||', Value("%", output_field=models.CharField())),
+            output_field=models.BooleanField()
+        )
+
+
 class DatabaseMaxCounter(AbstractCounter):
     """
     ``DatabaseMaxCounter`` is used to keep track of what data this database already has across all
@@ -593,8 +607,8 @@ class DatabaseMaxCounter(AbstractCounter):
         per_filter_max = []
 
         for filt in filters:
-            qs = queryset.annotate(filt=Value(filt, output_field=models.CharField()))
-            qs = qs.filter(filt__startswith=F("partition"))
+            qs = queryset.annotate(filter_matches=ValueStartsWithField(filt, "partition"))
+            qs = qs.filter(filter_matches=True)
             maxes = qs.values("instance_id").annotate(maxval=Max("counter"))
             per_filter_max.append({dmc["instance_id"]: dmc["maxval"] for dmc in maxes})
 ```

**Review by bjester — CHANGES_REQUESTED** [review]
I would prefer a beta release if we're unsure if this is a complete fix, or at least ensuring that the Django generated SQL does not contain the unnecessary bits it adds (with the code I shared)

File: morango/models/core.py (hunk 1 above)
Comment (jamalex) [reply morango/models/core.py]: That looks like a great tweak -- and yeah, a beta sounds good for now (don't see any blockers for doing one on Morango, but don't think we ever have -- I'm assuming Github Actions won't care for building it, and then the name will be parsed by PyPi, so it should just be magic. Works for me -- thanks!

**Review by bjester — APPROVED** [review]
Looks good

File: morango/__init__.py [hunk 2]
```diff
@@ -3,4 +3,4 @@
 from __future__ import unicode_literals
 
 default_app_config = "morango.apps.MorangoConfig"
-__version__ = "0.6.6"
+__version__ = "0.6.6-beta"
```
jamalex: Do we need to use [PEP 440](https://www.python.org/dev/peps/pep-0440/#pre-releases) in order for PyPI to handle this properly?

Something like `0.6.6b0`?

bjester: Ah yes that's what we want. I looked at Kolibri and saw `-alpha` on the releases so went with that

rtibbles: Yes, Kolibri turns the version tuple into a PEP compliant version string.

PR #135, opened 2021-09-29
Update calculate_max_sqlite_variables to read compilation option

jamalex:
SQLite 3.32.0 and later increased the default value for the `MAX_VARIABLE_NUMBER` compilation option from 999 to 32766. This PR replaces our brute force calculation method with an approach that should get us the value directly and efficiently, and not hit the CTE error for older SQLite versions that caused us to [revert an old approach](https://github.com/learningequality/morango/commit/29fe2a151b7424394ecfab1d55d1649553b0c256#), by doing some clever looping proposed by @bjester.

### DIFF
--- a/morango/sync/backends/utils.py
+++ b/morango/sync/backends/utils.py
@@ -18,29 +18,31 @@ def load_backend(conn):
 @lru_cache(maxsize=1)
 def calculate_max_sqlite_variables():
     """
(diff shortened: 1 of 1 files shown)


bjester (approved):
Nice!
```
>>> from morango.sync.backends.utils import calculate_max_sqlite_variables
>>> calculate_max_sqlite_variables()
250000
```

PR #137, opened 2021-10-07
Fix legacy update server's records total

bjester:
## Summary
- Ensures we update `records_total` on the server side after queuing for push
- Without `records_total`, [legacy servers won't dequeue](https://github.com/learningequality/morango/blob/v0.5.6/morango/api/viewsets.py#L393) even though we pushed data

## TODO

## Reviewer guidance

## Issues addressed

## Documentation

### DIFF
--- a/morango/sync/operations.py
+++ b/morango/sync/operations.py
@@ -1127,7 +1127,18 @@ class LegacyNetworkQueueOperation(NetworkLegacyNoOpMixin, NetworkOperation):
     Without ASYNC_OPERATIONS capability, the server will perform queuing during initialization
     """
 
-    pass
+    def handle(self, context):
+        """
+        :type context: NetworkSessionContext
+        """
+        self._assert(ASYNC_OPERATIONS not in context.capabilities)
+
+        # When pushing, this should occur after queuing locally, so we need to update the server
+        # with how many records we've queued for the push
+        if context.is_push:
+            self.update_transfer_session(context, records_total=context.transfer_session.records_total)
+
+        return transfer_statuses.COMPLETED
 
 
 class NetworkQueueOperation(NetworkOperation):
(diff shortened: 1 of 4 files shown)


morango/sync/operations.py
```diff
@@ -1127,7 +1127,18 @@ class LegacyNetworkQueueOperation(NetworkLegacyNoOpMixin, NetworkOperation):
...
-    pass
+    def handle(self, context):
+        """
+        :type context: NetworkSessionContext
+        """
+        self._assert(ASYNC_OPERATIONS not in context.capabilities)
+
+        # When pushing, this should occur after queuing locally, so we need to update the server
+        # with how many records we've queued for the push
+        if context.is_push:
```
jamalex, line 1143: I'm guessing this is the case, but just to double-confirm: when it's a pull *from* here, the counts on the other side will be properly updated without this mechanism? In other words, what matters here is that it's a push, not that we're the sending side?

Also, relatedly, is there a mechanism in place such that this task is only executed on the client side, not the server side? I can't remember if that's handled via asserts or elsewhere.

jamalex (approved):
Looks good to me, as long as the answers to my questions there are "yep, that's handled", as I'm expecting is the case!

morango/sync/operations.py
bjester: > when it's a pull from here, the counts on the other side will be properly updated without this mechanism?

The old approach performed queuing during the creation of the transfer session, so when the server then returns it's response during initialization, the client side then sets the `records_total`. [That's handled here](https://github.com/learningequality/morango/blob/release-v0.6.x/morango/sync/operations.py#L1050), which I think I'll add a comment to. This definitely does feel a little weird to not have the symmetry of the pull case here as well, but I think it's just a result of the unique combination of transfer stages and old vs new behavior restricting where we can actually place this logic.

> Also, relatedly, is there a mechanism in place such that this task is only executed on the client side, not the server side?

Yes there's a slight subtlety with the `Network` operations, since they expect a `NetworkSessionContext` which is only ever created on the client side. So while there's no `is_producer` attribute, `is_push` is sufficient in this context.

PR #138, opened 2021-10-15
Support for custom instance info

### LINKED ISSUE #136: Update morango to allow adding additional instance context, for example to share the version of Kolibri that is syncing

bjester:
## Summary
- Adds new dict setting `MORANGO_INSTANCE_INFO`
- Updates serializer for `InstanceIDModel` to include custom instance info from new setting
- Updates info API endpoint to include custom instance info from new setting

## TODO

## Issues addressed
Resolves https://github.com/learningequality/morango/issues/136

### DIFF
--- a/morango/constants/settings.py
+++ b/morango/constants/settings.py
@@ -2,6 +2,7 @@ ALLOW_CERTIFICATE_PUSHING = False
 MORANGO_SERIALIZE_BEFORE_QUEUING = True
 MORANGO_DESERIALIZE_AFTER_DEQUEUING = True
 MORANGO_DISALLOW_ASYNC_OPERATIONS = False
+MORANGO_INSTANCE_INFO = {}
 MORANGO_INITIALIZE_OPERATIONS = (
     "morango.sync.operations:InitializeOperation",
     "morango.sync.operations:LegacyNetworkInitializeOperation",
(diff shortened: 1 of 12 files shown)


morango/constants/settings.py
```diff
@@ -2,6 +2,7 @@
 MORANGO_SERIALIZE_BEFORE_QUEUING = True
 MORANGO_DESERIALIZE_AFTER_DEQUEUING = True
 MORANGO_DISALLOW_ASYNC_OPERATIONS = False
+MORANGO_INSTANCE_INFO = {}
```
rtibbles, line 5: Did you decide against making it an import path and just having a bare dictionary here?

rtibbles (commented):
Looks like it will work, but one question about implementation re our previous discussion.

morango/constants/settings.py
bjester: I didn't specifically decide against that. This works with Django's lazy object utilities and I think I recall you also mentioning that, but perhaps you were referring to the import path approach. I'm not sure which is more preferable, or more Django-style. If I had to guess, I'd bet using a `SimpleLazyObject` would use more memory than an import path. Thoughts?

rtibbles: I think defining an import path would be better - because if we just did a lazy object, we'd then have to do the imports in side a functional scope in order to prevent premature import evaluation in the settings. Also seems very Djangoic :)

rtibbles (approved):

PR #142, opened 2021-11-19
Also regenerate source ids when empty string and strictly enforce calculation

### LINKED ISSUE #141: Fix tests so we can re-merge #115
We had to back out https://github.com/learningequality/morango/pull/115 because the tests aren't passing yet. Once the tests are fixed (removing the "calling calculate_uuid will regenerate it even if already set" assumption), the changes can be reintroduced.

bjester:
## Summary
- Allows empty string to trigger source id regeneration
- Raises an error if `calculate_source_id` returns an empty string which is undefined behavior
- Fixes test issues and deprecation warnings

## TODO

## Issues addressed

Resolves https://github.com/learningequality/morango/issues/141

## Documentation

### DIFF
--- a/morango/api/urls.py
+++ b/morango/api/urls.py
@@ -10,16 +10,16 @@ from .viewsets import SyncSessionViewSet
 from .viewsets import TransferSessionViewSet
 
 router = routers.SimpleRouter()
-router.register(r"certificates", CertificateViewSet, base_name="certificates")
+router.register(r"certificates", CertificateViewSet, basename="certificates")
 router.register(
-    r"certificatechain", CertificateChainViewSet, base_name="certificatechain"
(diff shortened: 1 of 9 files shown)


jamalex (approved):
Looks good from my side!

Some of these changes (renamings) look like they rely on minimum versions of Django and/or DRF. Are they in line with what we have in the requirements?

bjester:
@jamalex Good call, the renaming requires bumping the minimum django version to `1.10`. Although, we could possibly bump the maximum version too since the changes handle deprecations in 2.x

jamalex:
Yeah, I guess we haven't tested other parts of the code in 2.x, though, so it may be safest not to bump the max (until we have reason to test).

PR #143, opened 2022-01-13
Make the transfer status fields nullable

bjester:
## Summary
- Prevents Django setting `""` default for new transfer status related fields

## TODO

- [x] Have tests been updated for the new code?

## Reviewer guidance

Tested both migration pathways, with previous migration already applied and without any migrations

## Issues addressed

## Documentation


jamalex (approved):
Testing both upgrades pathways, which you did, was my main question. Besides that, I couldn't see anything from a skim of the rest of the code the seemed to rely on these fields being `str` instances (mostly truthiness or equivalence), and the tests all pass, so looks good to me! Thanks!

PR #145, opened 2022-01-24
First pass implementation of FSIC v2 format, to fix #144

### LINKED ISSUE #144: Current FSIC format is not expressive enough to support full needs of single-user syncing in Kolibri
## Background

We discovered an issue in Kolibri with syncing from a single-user device A to a full-facility device B, and then from there to another full-facility device C, in which the data from A does not make it onto C.

The issue turned out to be with how we compute, communicate, and use the "FSIC" (Filter-Specific Instance Counter) data structure in Morango (also referred to as ["Filter Max Counters" in the docs](https://morango.readthedocs.io/en/latest/counters/index.html#counters)).

The current format for the FSIC is a dict of the form `{instance_id: counter}`, for the specific filter (set of partition prefixes) being used in the sync, mapping each `instance_id` to a `counter` level. Each side of the sync generates its own FSIC, which basically says "For data matching this filter, I have everything from these devices (instances) up through these counter levels."

## Problem

Here is a high-level cartoonish summary of how this format prevented the syncing of A's data from B to C. (Note that to simplify I'll ignore counters and just say "has the latest data created on", since it doesn't materially impact the current problem. And when I say "sync data for the full facility" or "sync data for user X", those are referring to the filter being used in that sync.)
- To start with, each of the devices has all the data that it created itself ("A has all the latest data created on A", etc).
- When A syncs data for user X to B, A says to B "I have all the data for user X created on A" and B says "I have all the data for user X created on B, but not that was created on A, so give me all the data for user X created on A". After that sync, B knows that it has all the data for user X created on A, and still has all data for the full facility created on B (itself).
- When B then syncs data for the full facility to C, both the data created on B and the additional data for user X created on A _should_ be included within that (and B knows that it _has_ it), but when B tells C what data it has for the full-facility filter, it's only currently able to say "I have the latest data created on B for the full facility", because it can't say with confidence that it also has ALL the data created on A _for the full facility_, because there might have been additional data created on A (e.g. for a user Y) that B didn't receive during the single-user sync for user X.

## Solution (in-progress)

The new FSIC format basically allows B to tell C "I have the latest data created on B for the full facility, and in addition I also have the latest data for user X created on A", which then allows them to negotiate what pieces need to be sent along, and have the data from A be included as part of that.

At a more technical level, the new FSIC format is of the form `{"super": {partition: {instance_id: counter}}, "sub": {partition: {instance_id: counter}}}`. The `partition` roughly corresponds to the "for user X" or "for the full facility" from the example above -- so, it's now able to say that it has everything under a specific partition from a specific device, without saying that it has _all_ the data from that device matching the current sync filter.

The `"sub"` key contains the partitions that are under the current sync filter, whereas the `"super"` key contains partitions that are prefixes of the filter. This allows us to use space as efficiently as possible, as we mark the counters for an instance at the highest-level partition, removing redundant counters on partitions that are contained within that partition (i.e. for which it is a prefix). The other side is then able to "reinflate" this structure into a format usable for filtering Store records for queuing, by propagating the counters down from superpartitions onto the subpartitions.

Things are looking good with the first-pass implementation of this new format (which will be enabled through a "capabilities" flag, so that new versions can still talk to old versions, though with the single-user syncing constraints noted above).

jamalex:
## Summary

This is still a work in progress, but the main chunks of the new code are in place for addressing #144.

The main things still needing doing:
- [x] Determining how to link up the negotiated capabilities info (intersection between caps on server/client) and the places where FSICs are generated and processed, for seamless support for both the old and new formats as appropriate to context.
- [x] Updating the tests that are still hardcoded to assume the old format (to set the flags to make them work), and adding a few more that hardcode the new format.

## TODO

## Reviewer guidance

No need to look through this yet (as-is), but I wanted to get the draft up so I could link to specific pieces.

## Issues addressed

Fixes #144


jamalex:
Pushed some further updates. Closer to having all tests pass now -- I'm not quite sure how to diagnose the remaining 4 tests (under `tests/testapp/tests/integration/test_syncsession.py`), as the tracebacks locally weren't very informative. Advice welcome!

I commented out the "very_very_very_very_many_fsics" tests for now using the new queuing code, as they did hit the max tree depth error with anything more than 995 instances. May still be good to add some chunking in to be safe?

morango/models/core.py
```diff
@@ -572,97 +582,146 @@ class Meta:
...
+            super_fsics = {}
+            for part in super_partitions:
+                super_fsics[part] = cls.get_instance_counters_for_partition(
+                    part, only_if_in_store=only_if_in_store
+                )
+            sub_fsics = {}
+            for part in sub_partitions:
+                sub_fsics[part] = cls.get_instance_counters_for_partition(
+                    part, only_if_in_store=only_if_in_store
+                )
```
bjester: For the above two super/sub steps that gather the instance counters while `only_if_in_store=True`, I think we can achieve a speedup by reducing them to a single query for each super/sub case. By structuring the query in a certain way, we can query for multiple partitions in a single query and that should scale significantly better than O(n), in contrast with this. 

The original code produces a query like:
```sql
SELECT morango_databasemaxcounter.instance_id, morango_databasemaxcounter.counter
FROM morango_databasemaxcounter
WHERE (
    morango_databasemaxcounter.partition = '5154da78c739fc2f678a9a549b1dc565'
    AND morango_databasemaxcounter.instance_id IN (
        SELECT DISTINCT U0.last_saved_instance AS Col1
        FROM morango_store U0
        WHERE U0.partition LIKE '5154da78c739fc2f678a9a549b1dc565%'
    )
);
```
and t takes roughly 17 seconds to execute on KDP. We should expect that to scale somewhat linearly with the number of partitions, so with a 60 second timeout we can query for ~3.5 partitions. If I understand correctly, this would be more of an issue with SoUDs and the "sub" partition case, since there would be distinct "partitions" to query for each SoUD?

An alternate approach to writing the same query uses `EXISTS` and we get similar performance:
```sql
SELECT morango_databasemaxcounter.instance_id, morango_databasemaxcounter.counter
FROM morango_databasemaxcounter
WHERE (
    morango_databasemaxcounter.partition = '5154da78c739fc2f678a9a549b1dc565'
    AND EXISTS(
        SELECT 1
        FROM morango_store U0
        WHERE (
            U0.partition LIKE morango_databasemaxcounter.partition || '%'
            AND U0.last_saved_instance = (morango_databasemaxcounter.instance_id)
        )
    )
);
```
This works in SQLite and Postgres. It's slightly slower in SQLite, but not enough to warrant attention.

With that change, we can further alter it to query for multiple at once:
```sql
SELECT morango_databasemaxcounter.partition, morango_databasemaxcounter.instance_id, morango_databasemaxcounter.counter
FROM morango_databasemaxcounter
WHERE (
    morango_databasemaxcounter.partition IN ('5154da78c739fc2f678a9a549b1dc565')
    AND EXISTS(                            
        SELECT 1                           
        FROM morango_store U0
        WHERE (
            U0.partition LIKE morango_databasemaxcounter.partition || '%'
            AND U0.last_saved_instance = (morango_databasemaxcounter.instance_id)
        )                    
    )          
);
```
The performance of that query slows to ~23 seconds, but querying with 2-3 partitions doesn't increase the query time, so when there are 3 partitions to query for, we get a 50% improvement over the original. With more partitions, we'll see even better performance in comparison. The downside is that there's a bug in Django that adds `= TRUE` to the `EXISTS(...)` clause when generating a query like this. It's fixed in later versions of Django but it causes postgres (9.6 at least) to choose poorer query plan even though it actually uses indices in that case.

Overall, I think we should avoid looping on this expensive query. Happy to help figure out how to make Django happy.

jamalex: Brilliant, thanks for the detailed analysis and benchmarking on this!

>If I understand correctly, this would be more of an issue with SoUDs and the "sub" partition case, since there would be distinct "partitions" to query for each SoUD?

That's correct -- there's typically one new subpartition added (with instance/counters specific to the SoUD, so they aren't redundant with a superpartition) per user who syncs from a SoUD. So parallelizing into a single query like that is definitely desirable.

>This works in SQLite and Postgres. It's slightly slower in SQLite, but not enough to warrant attention.

Just out of curiosity -- slower than Postgres, or also slower than the original query?

>Overall, I think we should avoid looping on this expensive query. Happy to help figure out how to make Django happy.

Thanks, would love your help with this!

A further thought: there may be even more ways to further consolidate, given that the list of partitions we're looping over there are themselves the products of applying the filters (of which there are generally very few) against the partitions in the DMC table. We'd still want a separate query for super vs sub, but other than that it should be possible. I kept things more separate for now given I was wanting to be able to easily introspect and reason about the inputs/outputs of each stage (and toggle on/off certain things), while finalizing the algorithm, but given it seems to be working, we could look towards further consolidation into some larger queries. Except if it doesn't buy us too much in terms of performance, it may not be worth it for now?

bjester: > Just out of curiosity -- slower than Postgres, or also slower than the original query?

The original query took 2-3ms on SQLite for the VPN DB, and the query using `EXISTS` took 6-7ms. So it's twice as slow, but also 3-5ms isn't very much.

> Except if it doesn't buy us too much in terms of performance, it may not be worth it for now?

Since we're already dealing with timeouts on KDP, I think we should try to avoid making that worse. Some small tweaks to avoid repeating expensive queries are some quick wins we can make to mitigate it. Like you said, with tests in place and passing, it's easy to swap out bits of code and verify the algorithm isn't altered negatively. Here's what makes Django happy and gets around the `EXISTS(...) = TRUE` issue:

```python
    @classmethod
    def get_instance_counters_for_partitions(cls, partitions, only_if_in_store=True):
        queryset = cls.objects.filter(partition__in=partitions)
        if only_if_in_store:
            matching_store_instances = (
                Store.objects.filter(
                    partition__startswith=models.OuterRef("partition"),
                    last_saved_instance=models.OuterRef("instance_id")
                )
                .values("last_saved_instance")
            )
            # manually add EXISTS clause to avoid Django bug which puts `= TRUE` comparison that kills postgres perf
            exists = models.Exists(matching_store_instances)\
                .resolve_expression(query=queryset.query, allow_joins=True, reuse=None)
            queryset.query.where.add(exists, "AND")

        queryset = queryset.values_list("partition", "instance_id", "counter")

        counters = {}
        for partition, instance_id, counter in queryset:
            if partition not in counters:
                counters[partition] = {}
            counters[partition].update({instance_id: counter})
        return counters
```

jamalex: Thanks so much! I'll give that a shot -- there are a couple more regressions tests I'm thinking to add now as well (to cover some bases with the `only_if_in_store` piece), just to be more fully confident on having that as the default.

>Since we're already dealing with timeouts on KDP, I think we should try to avoid making that worse. Some small tweaks to avoid repeating expensive queries are some quick wins we can make to mitigate it.

Sorry, I wasn't being very clear: I very definitely agree that avoiding the duplicated queries as you're doing here is critical. I was referring there to my suggestion that we could go even further and build more of a mega-query that didn't require us to first query for the list of partitions and then pass them in to this function. But I think we can work towards that part incrementally -- this piece here is the critical lower-hanging fruit.

jamalex:
I made an [additional tweak](https://github.com/learningequality/morango/pull/145/commits/ad3aff92ded811820ca48806e30e2e9b2c8fde97) to address an edge case for a setup with repeated one-directional transfers (i.e. just pushes, or just pulls, not bidirectional) in which the some of the same records might be sent again and again each time, even when the other side already has them. This was a result of the `only_in_store` stuff, which only looked at the `last_saved_instance` IDs, and not the `instance_id` in `RecordMaxCounter`. Now, we reintroduce the `is_producer` asymmetry, with the producer referencing the `Store` to determine what it actually has available to send, vs the receiver referencing `RecordMaxCounter` so it's clear when it already has a record even if the `last_saved_instance` on its local (newer) version of the record now has a different instance ID. I added a regression test, and then got it to go from failing to passing.

(In practice, this shouldn't currently have made a difference in Kolibri, given we always do both a push and pull, but I wanted to make sure we were robust to future scenarios or other use cases.)

jamalex:
The main thing that still feels important to consider with this is the union-based chunking of the queries, to handle when there are > ~1000 partitions and/or instances. I bumped the version to 0.6.8 alpha so we could start testing more in Kolibri in the meantime, though.

tests/testapp/tests/sync/test_operations.py
```diff
@@ -271,6 +271,320 @@ def test_local_queue_operation__noop(self, mock_queue):
...
+    #     """
+    #     fsics = {"super": {}, "sub": {"": {self.data["group1_id"].id: 1, self.data["group2_id"].id: 1}}}
+    #     fsics["sub"][""].update({uuid.uuid4().hex: i for i in range(20000)})
+    #     self.transfer_session.client_fsic = json.dumps(fsics)
+    #     self.transfer_session.server_fsic = json.dumps({"super": {}, "sub": {}})
+    #     _queue_into_buffer_v2(self.transfer_session)
+    #     # ensure all store and buffer records are buffered
+    #     assertRecordsBuffered(self.data["group1_c1"])
+    #     assertRecordsBuffered(self.data["group1_c2"])
+    #     assertRecordsBuffered(self.data["group2_c1"])
```
bjester, line 316: Are these still needed? What were the results of testing how many fsics we can allow?

jamalex: Oh, I thought I had posted this somewhere but I think it seems I probably hadn't: In testing the limits here, it started failing I think around 996 instance_ids. May be sufficient in practice for now, but not a comfortable limit to have longer-term.

Ideally I'd like to re-introduce chunking and uncomment this test before we go from an alpha to a release, but I was thinking in the interim (and perhaps sufficient even for the S2S pre-release) we could get this alpha integrated into Kolibri for further testing.

bjester (approved):
Looks good!

PR #146, opened 2022-02-05
Add further FSIC fix, update tests, and bump to a1

jamalex:
## Summary

Fixes a case where data that we have through a superpartition isn't included in a sync, by ensuring that we have all filter partitions on the FSIC explicitly. Thread here: https://learningequality.slack.com/archives/CB37UM23A/p1644025841936929?thread_ts=1643916096.314019&cid=CB37UM23A

## TODO

- [x] Have tests been written for the new code?
- [ ] Has documentation been written/updated?
- [ ] New dependencies (if any) added to requirements file

PR #147, opened 2022-02-07
Add chunking to FSIC v2 queuing

jamalex:
## Summary

When implementing `_queue_into_buffer_v2` for queuing records with the new FSIC v2, the first version didn't include chunking (and then UNIONing) of SELECT statements, to avoid "Expression tree is too large" errors. This PR builds on https://github.com/learningequality/morango/pull/146 (which should be merged first, to make this one easier to review) by adding back chunking logic.

## TODO

## Reviewer guidance

It basically splits the expanded FSIC into chunks with max partitions + instances of `chunk_size`, and then does the same logic as before to create a SELECT statement for each FSIC chunk, then does a UNION on those as in the old chunking approach.

It also makes a tweak to the chunking code in the v1 queuing operation, after noting that we don't need to create a separate SELECT and then UNION for the RMBC queries, since they were all identical.

### DIFF
--- a/morango/sync/operations.py
+++ b/morango/sync/operations.py
@@ -545,12 +545,12 @@ def _queue_into_buffer_v2(transfersession):
     assert "super" in client_fsic
 
     # ensure that the partitions in the FSICs are under the current filter, before using them
-    for partition in itertools.chain(server_fsic["sub"].keys(), client_fsic["sub"]):
+    for partition in itertools.chain(server_fsic["sub"].keys(), client_fsic["sub"].keys()):
         if partition not in sync_filter:
             raise MorangoInvalidFSICPartition("Partition '{}' is not in filter".format(partition))
 
-    server_fsic = expand_fsic_for_use(server_fsic)
-    client_fsic = expand_fsic_for_use(client_fsic)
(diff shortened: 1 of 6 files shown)


morango/sync/operations.py
```diff
@@ -580,39 +604,41 @@ def _queue_into_buffer_v2(transfersession):
...
-            store=Store._meta.db_table,
+        # combine conditions and filter by profile
+        where_condition = _join_with_logical_operator(
+            profile_condition + partition_conditions, "AND"
         )
-    )
+
+        # execute raw sql to take all records that match condition, to be put into buffer for transfer
+        select_buffers.append(
+            """SELECT DISTINCT
```
bjester: The `DISTINCT` shouldn't be needed here, but it shouldn't actually do anything either.

jamalex: Good point. Duplication is more likely to occur _across_ SELECTs, not _within_ one of them, but even there we're probably safe now that we remove redundancy in the FSIC. I'll remove it.

jamalex: Removed -- let's see what the tests think of that

bjester (approved):
Looks good!

PR #150, opened 2022-02-17
Add ORDER BY to buffer's API viewset

### LINKED ISSUE #148: Buffers API does not use ORDER BY in query expression
## Observed behavior
A database constraint error was encountered while pulling a large amount of records from KDP. As you can see in the logs below, the process was in the middle of a sync when it encountered the error. Since both sides have the unique constraint, we know the data in the database must be unique, so the issue lies with transfer mechanism of the API.
```
INFO 2022-02-16 21:01:22,508 kolibri.core.auth.management.utils Receiving data (14400/17233, 23.29MB)
INFO 2022-02-16 21:01:24,518 kolibri.core.auth.management.utils Receiving data (14600/17233, 23.65MB)
ERROR 2022-02-16 21:01:24,531 kolibri.core.tasks.worker Job c5e1312596704cc2a1d6ab4bb50d0fe8 raised an exception: Traceback (most recent call last):
  File "/home/bjester/.pex/install/kolibri-0.15.1b0-py2.py3-none-any.whl.24346e050dea30c1152ccb24cde19a2d5c90ca1c/kolibri-0.15.1b0-py2.py3-none-any.whl/kolibri/dist/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/home/bjester/.pex/install/kolibri-0.15.1b0-py2.py3-none-any.whl.24346e050dea30c1152ccb24cde19a2d5c90ca1c/kolibri-0.15.1b0-py2.py3-none-any.whl/kolibri/dist/django/db/backends/sqlite3/base.py", line 328, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.IntegrityError: UNIQUE constraint failed: morango_buffer.transfer_session_id, morango_buffer.model_uuid
```

The buffers API, which supports `LIMIT`/`OFFSET` pagination, does not add an `ORDER BY` clause to the database query. In some databases, this may not matter. In PostgreSQL, selecting subsets of the query's with `LIMIT`/`OFFSET` can return duplicates between queries. This is because without an `ORDER BY`, the order used in the sub-selection of records is undefined, and thus may be inconsistent between repeated queries. This is what the [PostgreSQL documentation](https://www.postgresql.org/docs/9.6/queries-limit.html#:~:text=When%20using%20LIMIT,specified%20ORDER%20BY.) says:
> When using LIMIT, it is important to use an ORDER BY clause that constrains the result rows into a unique order. Otherwise you will get an unpredictable subset of the query's rows. You might be asking for the tenth through twentieth rows, but tenth through twentieth in what ordering? The ordering is unknown, unless you specified ORDER BY.

## Desired behavior
The buffers API should properly add an `ORDER BY` clause when using `LIMIT`/`OFFSET` so that all records are transferred without encountering duplication issues.

bjester:
## Summary
- Adds regression test for the `ORDER BY` issue on the buffers API
- Adds `docker-compose` config and updates `Makefile` to support running all tests with postgres using easy single command
- Buffers during push sync are already ordered correctly

## TODO

## Reviewer guidance
- See [this PR](https://github.com/learningequality/morango/pull/149) for failing test prior to changes.
- To run tests with Postgres: `make test-with-postgres` (must have `docker-compose`)

## Issues addressed
Resolves https://github.com/learningequality/morango/issues/148

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 List of the most important changes for each release.
 
+## 0.6.9
+- Fixes un-ordered selection of buffers during sync which can allow duplicates to be synced with PostgreSQL backends
+
 ## 0.6.8
 - Fixes subset syncing issues by introducing new FSIC v2 format
 
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+SHELL := /bin/bash
+
+.ONESHELL:
 .PHONY: help clean clean-pyc release dist
 
 define BROWSER_PYSCRIPT
@@ -23,11 +26,18 @@ endef
 export PRINT_HELP_PYSCRIPT
 BROWSER := python -c "$$BROWSER_PYSCRIPT"
 
+export PYTHONPATH
+export DJANGO_SETTINGS_MODULE
+
 help:
 	@echo "clean-build - remove build artifacts"
 	@echo "clean-pyc - remove Python file artifacts"
 	@echo "release - package and upload a release"
 	@echo "dist - package"
+	@echo "test - run python tests"
+	@echo "test-with-postgres - run python tests with docker postgres backend"
+	@echo "tox - run all tests, with existing postgres backend"
+	@echo "tox-with-postgres - run all tests, with docker postgres backend"
 
 clean: clean-build clean-pyc clean-test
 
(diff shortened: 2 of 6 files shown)


jamalex (approved):
The Python changes look good to me. I defer to you and @rtibbles on the Makefile/build/testrunner pieces, but that feels low-risk from my side too.

PR #152, opened 2022-02-23
Move updating of FSICS to after dequeuing

### LINKED ISSUE #151: DatabaseMaxCounters are not updated after dequeuing, but after deserialization
## Background
Prior to the new sync operations architecture, deserialization occurred within the same database transaction in which dequeuing also occurred. So if there was a failure during deserialization, any changes to the `Store` should have been rolled back and thus database counters would continue to match the current state of the `Store`.

With the new architecture, updating the counters was placed after deserialization, because that kept it consistent in where it occurred previously with regards to sync timeline. Although, since the new operations architecture separated dequeuing and deserialization, if deserialization failed, the database counters would not reflect the changes to the `Store` made during dequeuing.

## Observed behavior
Deserialization conflicts have arisen in the data handling on KDP due to this issue. 

## Expected behavior
The database counters are updated whenever the `Store` is updated. This also has the benefit that should the sync be interrupted during the dequeuing stage, a retry of the sync should determine that it doesn't need to resync the same data, even though deserialization wasn't completed.

bjester:
Requires https://github.com/learningequality/morango/pull/150, so that I can update the changelog with this change in this same PR

## Summary
- Fixes https://github.com/learningequality/morango/issues/151

## TODO

### DIFF
--- a/morango/sync/operations.py
+++ b/morango/sync/operations.py
@@ -665,7 +665,7 @@ def _queue_into_buffer_v2(transfersession, chunk_size=200):
 
 
 @transaction.atomic(using=USING_DB)
-def _dequeue_into_store(transfersession):
+def _dequeue_into_store(transfer_session, fsic, v2_format=False):
     """
     Takes data from the buffers and merges into the store and record max counters.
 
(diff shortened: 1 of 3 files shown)


morango/sync/operations.py
```diff
@@ -968,6 +968,18 @@ def handle(self, context):
...
+            context.transfer_session.client_fsic
+            if context.is_server
+            else context.transfer_session.server_fsic
+        )
+        DatabaseMaxCounter.update_fsics(
+            json.loads(fsic),
+            context.filter,
+            v2_format=FSIC_V2_FORMAT in context.capabilities,
+        )
+
```
jamalex, line 986: We'll want to ensure this happens within the same transaction as the dequeuing itself, since otherwise the database could still end up in an inconsistent state. I think moving this inside `_dequeue_into_store` would be sufficient, given that function is wrapped with `@transaction.atomic(using=USING_DB)`.

jamalex (approved):
I suggested moving one block of code to ensure transactionality, but otherwise looks great to me -- thanks!

morango/sync/operations.py
jamalex: (we also don't need to worry about needing this to be run if no records were transferred)

jamalex (approved):
Looks great, thanks!

PR #154, opened 2022-03-07
Fix Postgres issue not dropping null constraint

bjester:
## Background
In https://github.com/learningequality/morango/pull/143 we revised an existing migration which added new columns to make the fields nullable. In addition, we created another migration that also drops the null constraint if the previous migration had already ran. This appears to work fine for SQLite but not for PostgreSQL due to something internal to Django (1.11 at least which is used in [Kolibri](https://github.com/learningequality/kolibri/issues/9139)).

## Summary
- Resolves issues with modified migrations that don't apply correctly with Postgres

## TODO

## Reviewer guidance
- See postgres test failures [here](https://github.com/learningequality/morango/runs/5457294906?check_suite_focus=true) before changing the `migrate_to` of the test to the new migration

## Issues addressed

https://github.com/learningequality/kolibri/issues/9139

## Documentation


rtibbles (approved):
Code changes make sense, test covers it.

PR #159, opened 2022-04-14
Defer foreign key checks for bulk processing during deserialization

### LINKED ISSUE #111: Deserialization validation logic repeats queries
## Background
Deserialization can be a lengthy process.

## Observed behavior
It appears that there are a significant number of repeated and unnecessary database queries that occur while validating 
 and deserializing transferred data.

## Desired behavior
Validation logic should be optimized such that it shouldn't produce repeated queries, thereby ensuring that Morango operations execute as swiftly as possible.

bjester:
## Summary
Adds bulk processing of foreign keys during deserialization to avoid excessively querying the database to validate them.

## TODO

## Reviewer guidance
- Do the changes make sense?
- Testing sufficient?
- Any other performance improvements that can be made?

## Issues addressed

Fixes https://github.com/learningequality/morango/issues/111


morango/sync/backends/utils.py
```diff
@@ -46,3 +50,102 @@ def calculate_max_sqlite_variables():
...
+        for value_dict in values:
+            for field in self.fields:
+                params.append(value_dict.get(field.attname))
+        with self.connection.cursor() as c:
+            self.backend._bulk_insert(c, self.sql_name, self.fields, params)
+
+    class Meta:
+        """
+        HACK: some Django code bits require a model, only to access meta information, which we use
+        this to mimic that on this non-model class
```
jamalex: Could remove "that" (or change "which" to "and" or "so" in previous line)

jamalex (approved):
This looks great! I could never have gotten 5% of the way there on this -- super powerful approach!

My one "nice to have" would be to have the deserialization errors that are written to the Store model be in the same format as they used to be (I believe the text may be different now), as that would make future querying for stuff that won't validate easier, rather than a mix of formats. But happy to merge and deploy as-is, also!

tests/testapp/tests/sync/test_controller.py
```diff
@@ -641,7 +641,7 @@ def test_deserialization_of_model_with_missing_foreignkey_referent(self):
 
         new_log.refresh_from_db()
         self.assertTrue(new_log.dirty_bit)
-        self.assertIn("missing MyUser", new_log.deserialization_error)
+        self.assertIn("my user instance with id '{}'".format(data["user_id"]), new_log.deserialization_error)
```
jamalex, line 644: Where does the space come from in between "my" and "user" here?

jamalex: Oh, I see the regex above. I think keeping it as the PascalCase class name would be more consistent with the existing stuff.

bjester: In the dict format examples, the model name appeared with spaces, like `mastery log`. So I added a regex that turns the pascal case model name into this format. In this case, the model's name is `MyUser`.

jamalex: Ah yes, let's keep that then! 👍

PR #162, opened 2022-04-27
Add an index to the Store partition field

### LINKED ISSUE #160: Add index to the Store table's partition field
## Background
We have no indices on the `Store` (`morango_store`) that benefit the queuing process. On postgres instances where this table is significantly large, that can cause lengthy sequential scans on the entire table.

## Deliverables
- Add an index to `Store.partition`
- The index should be applied `CONCURRENTLY` on Postgres
- The index should allow `text_pattern_ops` on Postgres
- The index should be created only if it doesn't exist (some instances may have had this added manually)

rtibbles:
## Summary

* Adds an index on the Store partition field
* For every backend except postgres just uses a regular AddIndex migration
* For Postgres, adds the index `CONCURRENTLY` and using the `IF NOT EXISTS` condition
* Also for Postgres uses the `text_pattern_ops` operator to ensure it is most useful for indexing a text field

## TODO

## Reviewer guidance

I've added a test for the migration specifically on postgres, but only to test the IF NOT EXISTS behaviour. I think just the fact that the migration runs at all might be sufficient for a test otherwise?

## Issues addressed

Fixes #160


bjester (approved):
:100:

PR #163, opened 2022-04-28
Just modify the Django UUIDField for our 32 digit hex representation.

### LINKED ISSUE #89: TypeErrors on UUID validation actually result in KeyErrors
https://github.com/learningequality/morango/blob/master/morango/models/fields/uuids.py#L40

This line references the `invalid` message on the `self.error_messages` dict, which seems not to have this key (at least not all the time). This causes a `KeyError` to be raised, leading to uncaught errors during field validation.

rtibbles:
## Summary
* Reduces the amount of copied in code from Django by just inheriting from the Django UUIDField implementation
* Avoids errors caused by incomplete copying

## TODO

## Reviewer guidance

Don't think I've missed anything but I'd appreciate @jamalex's eyes to be sure.

I have run the morango integration tests in Kolibri using these changes installed, and everything passed.

## Issues addressed

Fixes #89

### DIFF
--- a/morango/models/fields/uuids.py
+++ b/morango/models/fields/uuids.py
@@ -10,60 +10,27 @@ def sha2_uuid(*args):
     return hashlib.sha256("::".join(args).encode("utf-8")).hexdigest()[:32]
 
 
-class UUIDField(models.CharField):
+class UUIDField(models.UUIDField):
     """
     Adaptation of Django's UUIDField, but with 32-char hex representation as Python representation rather than a UUID instance.
     """
-
-    def __init__(self, *args, **kwargs):
-        kwargs["max_length"] = 32
-        super(UUIDField, self).__init__(*args, **kwargs)
-
-    def prepare_value(self, value):
-        if isinstance(value, uuid.UUID):
-            return value.hex
-        return value
-
-    def deconstruct(self):
-        name, path, args, kwargs = super(UUIDField, self).deconstruct()
-        del kwargs["max_length"]
-        return name, path, args, kwargs
-
-    def get_internal_type(self):
-        return "UUIDField"
 
     def get_db_prep_value(self, value, connection, prepared=False):
         if value is None:
             return None
         if not isinstance(value, uuid.UUID):
-            try:
-                value = uuid.UUID(value)
-            except AttributeError:
-                raise TypeError(self.error_messages["invalid"] % {"value": value})
+            value = super(UUIDField, self).to_python(value)
+
+        if connection.features.has_native_uuid_field:
+            return value
         return value.hex
 
     def from_db_value(self, value, expression, connection):
         return self.to_python(value)
 
     def to_python(self, value):
-        if isinstance(value, uuid.UUID):
-            return value.hex
-        return value
-
-    def get_default(self):
-        """
-        Returns the default value for this field.
-        """
-        if self.has_default():
-            if callable(self.default):
-                default = self.default()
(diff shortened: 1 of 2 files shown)


morango/models/fields/uuids.py
```diff
@@ -2,67 +2,35 @@
...
     """
 
-    def __init__(self, *args, **kwargs):
-        kwargs["max_length"] = 32
-        super(UUIDField, self).__init__(*args, **kwargs)
-
-    def prepare_value(self, value):
-        if isinstance(value, uuid.UUID):
-            return value.hex
-        return value
```
jamalex, line 25: This one doesn't seem to be captured in the parent class.

jamalex: (and is an important part of having it return the 32 char value rather than a UUID object, from what I recall)

morango/models/fields/uuids.py
```diff
@@ -2,67 +2,35 @@
...
         if not isinstance(value, uuid.UUID):
-            try:
-                value = uuid.UUID(value)
-            except AttributeError:
-                raise TypeError(self.error_messages["invalid"] % {"value": value})
+            value = super(UUIDField, self).to_python(value)
+
+        if connection.features.has_native_uuid_field:
+            return value
         return value.hex
```
jamalex, line 26: I'm now not seeing a difference between this and the parent class implementation, other than the use of `super`, which doesn't seem necessary, since `to_python` isn't defined in this class, so it'll use the super's anyway. So I'm guessing this can be removed?

morango/models/fields/uuids.py
jamalex: (I think maybe specifically for cases where someone sets it on there as a UUID object and then reads it back, without it going through the database yet)

But it could probably call `to_python` here to be DRYer.

jamalex (changes requested):
Probably a good move overall, but a few notes/concerns.

morango/models/fields/uuids.py
rtibbles: It is, see below.

morango/models/fields/uuids.py
rtibbles: `prepare_value` is a form field method though, not a regular field: https://github.com/django/django/blob/stable/1.11.x/django/forms/fields.py#L129

I do wonder if that's where some of the original implementation errors came from, as it seems the original implementation here was a copy of the Django form field UUIDField: https://github.com/django/django/blob/stable/1.11.x/django/forms/fields.py#L1206 but then wrapped around the model CharField implementation instead of the form field.

rtibbles: There is definitely something missing though, because during the serialization tests, it is returning as a UUID.

rtibbles: The default Django serialization framework uses the `value_to_string` method, which we seem to have not defined here.

See: https://github.com/django/django/blob/stable/1.11.x/django/db/models/fields/__init__.py#L834

Although we're then not using that in the serialize method: https://github.com/learningequality/morango/blob/release-v0.6.x/morango/models/core.py#L939

rtibbles: Have now updated to add a `value_from_object` override that properly coerces the value to a 32 digit hex string.

jamalex (approved):
Looks good to me, once it's been tested within Kolibri's test suite as well!

bjester (approved):
All looks good to me

PR #164, opened 2022-04-29
Replace serializable transaction isolation with advisory locking

bjester:
## Summary
Prior to these changes, when Morango detected the database backend was postgres, the de/queuing and de/serialization operations opened a transaction on a separate DB connection with serializable transaction isolation. While it opened a transaction on this connection, most if not all of the code that generated queries did not use this same connection. As a result, it seems there's a very real possibility of data corruption since changes to the database aren't encapsulated within a DB transaction-- they aren't atomic. Although, I suspect that the prior implementation did still handle concurrent writes because the transaction acts like a lock within the database while the operations take place. Recent changes that added temporary tables made this issue abundantly clear on KDP because the temporary table was immediately dropped since it wasn't executed within a DB transaction. None of this was apparent in the tests because the tests usually run within a transaction anyways.

Instead of a connection with serializable transaction isolation, the work here implements advisory locking to lock affected partitions during these operations. We implemented advisory locking on Kolibri Studio to lock the channel tree during MPTT updates. This has several benefits since we'll have more control over how concurrency is actually handled. For instance, queuing buffers now open a shared lock, while all other operations open exclusive locks, allowing for more concurrency during pull syncs. And since the locking will be based off affected partitions, we'll have more capacity for concurrent syncs since syncs for separate partitions won't block each other as they would with serializable transaction isolation.

## TODO

## Reviewer guidance

## Issues addressed

## Documentation

### DIFF
--- a/morango/sync/backends/postgres.py
+++ b/morango/sync/backends/postgres.py
@@ -1,4 +1,5 @@
 import binascii
+import logging
 
 from .base import BaseSQLWrapper
 from .utils import get_pk_field
@@ -15,6 +16,8 @@ LOCK_PARTITION = 2
 
 SIGNED_MAX_INTEGER = 2147483647
 
+logger = logging.getLogger(__name__)
+
 
 class SQLWrapper(BaseSQLWrapper):
     backend = "postgresql"
@@ -22,10 +25,30 @@ class SQLWrapper(BaseSQLWrapper):
         "CREATE TEMP TABLE {name} ({fields}) ON COMMIT DROP"
     )
 
+    def _transaction_has_savepoint(self):
+        """
+        Determine if we're in a transaction and whether savepoints have been created during it
+        :return:
+        """
+        if not self.connection.in_atomic_block:
+            return False
+        for savepoint_id in self.connection.savepoint_ids:
+            if savepoint_id is not None:
+                return True
+        return False
+
     def _set_transaction_repeatable_read(self):
         """Set the current transaction isolation level"""
         from psycopg2.extensions import ISOLATION_LEVEL_REPEATABLE_READ
-        if not SETTINGS.MORANGO_TEST_POSTGRESQL:
+
+        # setting the transaction isolation must be either done at the BEGIN statement, or before
+        # any reading/writing operations have taken place, which includes creating savepoints
+        if self._transaction_has_savepoint():
+            # if we're running tests, we should simply ignore this warning, since the test suites
+            # manage their own connections
+            if not SETTINGS.MORANGO_TEST_POSTGRESQL:
+                logger.warning("Unable to set transaction isolation when savepoints have been created")
+        else:
             self.connection.connection.set_isolation_level(ISOLATION_LEVEL_REPEATABLE_READ)
 
     def _prepare_with_values(self, name, fields, db_values):
--- a/morango/sync/backends/postgres.py
+++ b/morango/sync/backends/postgres.py
@@ -324,6 +324,11 @@ class SQLWrapper(BaseSQLWrapper):
         :param partition: The partition prefix string to lock
         :param shared: Whether the lock is exclusive or shared
         """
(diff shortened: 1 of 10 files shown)


morango/sync/utils.py
```diff
@@ -230,3 +230,21 @@ def __exit__(self, *args, **kwargs):
...
+    :type backend: morango.sync.backends.base.BaseSQLWrapper
+    :type sync_filter: morango.models.certificates.Filter|None
+    :type shared: bool
+    """
+    if sync_filter is None:
+        backend._lock_all_partitions(shared=shared)
+    else:
+        # parse the unique partitions in the filter
+        partitions = set(f.split(":")[0] for f in sync_filter)
+        # for every unique partition, acquire a lock
```
jamalex, line 248: Partitions are prefix-matching, whereas the crc32-hashing-based lock ID will just be exact-matching, so this wouldn't prevent, for instance, a full facility sync from colliding with a single-user sync, as the latter uses different partitions, but they're all suffixes of the full facility partition (which is the dataset_id).

My recommendation would be to just grab a single lock, for the primary partition (the first 32 chars of the partition), as that is the lowest-common-denominator that will prevent syncs for the same facility from stepping on one another's toes, while allowing concurrency across facilities. I'm using Kolibri examples, but the same principles apply more generally across potential uses of Morango.

morango/sync/backends/postgres.py
```diff
@@ -268,3 +277,63 @@ def _dequeuing_insert_remaining_rmcb(self, cursor, transfersession_id):
...
+        keys = [key1]
+        if key2 is not None:
+            keys.append(key2)
+
+        query = "SELECT pg{_try}_advisory_{xact_}{lock}{_shared}({keys}) AS lock;".format(
+            _try="" if wait else "_try",
+            xact_="" if session else "xact_",
+            lock="unlock" if unlock else "lock",
+            _shared="_shared" if shared else "",
+            keys=", ".join(["%s" for i in range(0, 2 if key2 is not None else 1)])
```
jamalex: Took me a moment to parse the list comprehension here. Totally optional, but one alternative format could be something like `["%s"] * len(keys)`.

jamalex (changes requested):
Thanks! This is fantastic for isolating syncs from one another. I've never played with DB advisory locks, but they seem very powerful for this type of use case -- thanks for the TIL!

I made one note about selecting the partition to lock on, but a more fundamental thought: This looks like it should handle potential conflicts between multiple syncs, but the other side of this I forgot to mention but am remembering now is the need for snapshotting during serialization, where the app models may be getting modified in parallel through the app's API etc, while a serialization task is in progress. So without read-level isolation there, we may pull some records from app models and serialize them, then the app makes changes to both the models we already pulled and some we're about to pull, leading to inconsistencies in what we put into the Store. Is there a special transaction type we could use just for the serialization operation to ensure we're reading from a static snapshot over the course of that serialization, across all the tables we're looping over?

Related to that, and also happening during serialization, was the [incrementing of the current local instance counter](https://github.com/learningequality/morango/blob/7c497ec6401af50bb382902f068bd619af972b04/morango/models/core.py#L194), and ensuring it's an in-place increment rather than a "get and then set" operation that could have race conditions (and lead to two serializations using the same counter etc). I _think_ we may be OK with the current implementation, because it uses an `F` field reference for single-statement incrementing rather than a read/write, but if you think it would be safer, we could add an instance-level lock in there too.

jamalex (approved):
This looks great, thanks!

I saw that setting it as repeatable read can have memory implications. The main place it's relevant is in the serialization stage, where it's reading from the app models, which can be changed through non-Morango operations. For the Morango-internal operations, is it worth assessing whether the partition-based advisory locking fully covers our bases, in which case we may be able to skip the repeatable read on some of these operations? Or if the overhead is minimal, this may be safest anyway.

Might be good for us to do some stress testing with this once we have it on, say, KDP develop -- with a bunch of clients syncing data for the same facility, or for different ones?

PR #165, opened 2022-05-05
Limit transaction isolation to de/serialization and retry for serializable isolation errors

bjester:
## Summary

The _repeatable-read_ transaction isolation can raise errors when there have been concurrent updates to records modified during the transaction. This update catches them at the transfer stage operations level to return a `PENDING` status which signals that the operation should be retried, either for a status update (like for async operations) or to actually retry them (like in this case)

```
ERROR:  could not serialize access due to concurrent update
```

## TODO

### DIFF
--- a/morango/sync/backends/postgres.py
+++ b/morango/sync/backends/postgres.py
@@ -37,6 +37,20 @@ class SQLWrapper(BaseSQLWrapper):
                 return True
         return False
 
+    def _is_transaction_isolation_error(self, error):
+        """
+        Determine if an error is related to transaction isolation
+        :param error: An exception
+        :return: A bool whether the error is a transaction isolation error
+        """
+        from psycopg2.extensions import TransactionRollbackError
+
+        # Django can wrap errors, adding it to the `__cause__` attribute
+        for e in (error, getattr(error, '__cause__', None)):
+            if isinstance(e, TransactionRollbackError) and "concurrent update" in str(e):
+                return True
+        return False
+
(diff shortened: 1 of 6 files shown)


morango/sync/backends/postgres.py
```diff
@@ -37,6 +37,18 @@ def _transaction_has_savepoint(self):
...
 
+    def _is_transaction_isolation_error(self, error):
+        """
+        Determine if an error is related to transaction isolation
+        :param error: An exception
+        :return: A bool whether the error is a transaction isolation error
+        """
+        # Django can wrap errors, adding it to the `__cause__` attribute
+        for e in (error, getattr(error, '__cause__', None)):
+            if "could not serialize access due to concurrent" in str(e):
```
jamalex: Could this string potentially be internationalized, and hence not get matched here, on non-EN machines/configs?

Looks like at least some Postgres errors do get localized: https://stackoverflow.com/questions/2404261/how-to-disable-postgresql-messages-translation

jamalex (approved):
Looks good to me! Fancy concurrent test! 🥇 

Just had one question on there, and I'm not sure quite how we'd address it, but maybe we can just open up a tech debt issue for it if you think it might be a concern.

morango/sync/backends/postgres.py
bjester: Good catch! So I guess I can look for the TransactionRollback error. I was checking for that and the str, but removed that, so 3rd time's the charm

PR #167, opened 2022-05-23
Consolidate sync client operation to use composite context

### LINKED ISSUE #157: Local transfer session reports stage is completed but remote server is still processing
## Background
Morango provides two classes for clients to manage pull and push syncs, respectively. These classes manage two different contexts-- one for the local instance and one for the remote instance. This structure creates symmetry between operating on the local client instance and the remote server instance.

## Observed behavior
During a pull sync, while the remote instance is queuing, the state of the local transfer session is saved with the following before the remote instance has even finished queuing.
```json
{
  "transfer_stage": "queuing",
  "transfer_stage_status": "completed"
}
```
The reason for this is because with the two contexts, only one can be operated on at once, and the local context is operated on first. In this case, with the pull sync, there's nothing to queue.

## Desired behavior
The local transfer session state matches the remote state as closely as possible.

bjester:
## Summary
- Adds new `CompositeSessionContext` class that can manage multiple contexts at once, specifically for managing a local and remote context simultaneously, ensuring the state of each is transfer session, local and remote, is kept in sync as much as possible
- Refactors the sync client class to use a new context class that wraps the local and remote contexts
- Other minor refactoring to accommodate existing behavior

## TODO

## Issues addressed

https://github.com/learningequality/morango/issues/157

### DIFF
--- a/morango/sync/controller.py
+++ b/morango/sync/controller.py
@@ -186,7 +186,9 @@ class SessionController(object):
         # should always be a non-False status
         return result
 
-    def proceed_to_and_wait_for(self, target_stage, context=None, max_interval=5):
+    def proceed_to_and_wait_for(
+        self, target_stage, context=None, max_interval=None, callback=None
+    ):
         """
         Same as `proceed_to` but waits for a finished status to be returned by sleeping between
         calls to `proceed_to` if status is not complete
@@ -196,18 +198,24 @@ class SessionController(object):
         :param context: Override controller context, or provide it if missing
         :type context: morango.sync.context.SessionContext|None
         :param max_interval: The max time, in seconds, between repeat calls to `.proceed_to`
+        :param callback: A callable to invoke on every retry
         :return: transfer_status.* - The status of proceeding to that stage,
             which should be `ERRORED` or `COMPLETE`
         :rtype: str
         """
         result = transfer_statuses.PENDING
         tries = 0
+        context = context or self.context
+        max_interval = max_interval or context.max_backoff_interval
+
         while result not in transfer_statuses.FINISHED_STATES:
             if tries > 0:
                 # exponential backoff up to max_interval
                 sleep(min(0.3 * (2 ** tries - 1), max_interval))
             result = self.proceed_to(target_stage, context=context)
             tries += 1
+            if callable(callback):
+                callback()
         return result
 
     def _invoke_middleware(self, context, middleware):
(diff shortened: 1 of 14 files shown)


morango/sync/controller.py
```diff
@@ -196,18 +198,24 @@ def proceed_to_and_wait_for(self, target_stage, context=None, max_interval=5):
         :param context: Override controller context, or provide it if missing
         :type context: morango.sync.context.SessionContext|None
         :param max_interval: The max time, in seconds, between repeat calls to `.proceed_to`
+        :param callback: A callable to invoke on every retry
```
jamalex: Nitpick (optional): might be clearer to name it `retry_callback`, as otherwise I might initially assume it's called once the proceeding has completed.

morango/sync/controller.py
```diff
@@ -223,32 +231,38 @@ def _invoke_middleware(self, context, middleware):
         stage = middleware.related_stage
         signal = getattr(self.signals, stage)
         at_stage = context.stage == stage
+        primed_context = None
 
         try:
             context.update(stage=stage, stage_status=transfer_statuses.PENDING)
 
+            # we'll use the primed context for passing to the middleware and any signal handlers
+            primed_context = context.prime()
```
jamalex: I see what it's doing, but I'm still not 100% clear on the "prime" name here -- is that a concept from a particular framework? My familiarity with the verb is from "priming a pump" or "semantic priming", being something that prepares something to then happen more easily. So my initial assumption might be that this method would have side effects, rather than returning a child.

(Not a request to change, just working to wrap my head around some of the terminology/framework.)

jamalex (approved):
Looks great! I had a couple terminology questions in there, but nothing that requires changes. Thanks!

morango/sync/controller.py
bjester: I didn't really like having to add this `prime` method but it seemed like the simplest change to accommodate the composite context in the controller architecture. As far as the name, `prime`, it's the best that I came up with and didn't originate from any framework. If you have suggestions, I'd be happy to hear them.

morango/sync/controller.py
bjester: So I corrected the parameter comment, since the callback is always invoked after each attempt

morango/sync/controller.py
bjester: Renamed this to `prepare`

PR #169, opened 2022-06-15
Refactor transaction isolation handling and some cleanup

### LINKED ISSUE #170: Modifying the database connection's transaction isolation persists beyond the transaction
## Background
We recently made fixes to Morango that correct the database transaction handling, which was meant to perform certain operations within a transaction with higher transaction isolation when using postgres. Due to connection mismanagement, the operations weren't actually executing in a transaction.

## Observed behavior
A bad assumption was made by @bjester thinking that setting the transaction isolation level after opening a transaction would execute SQL that modifies the current transaction's isolation level. Instead, the `BEGIN` of the transaction is actually delayed until SQL is executed through the connection, which means the isolation is enabled through the `BEGIN` statement and persists for all transactions opened through that connection until it is closed.

## Expected behavior
Morango should apply the higher transaction isolation level for the duration of the transaction, by rolling it back to its previous level after the transaction has been closed, either by `ROLLBACK` or `COMMIT`.

bjester:
## Summary
- Adds explicit `on_delete=models.CASCADE` to all models since its implicit definition is deprecated in Django 2+
- Refactors modification of the transaction isolation level to be more robust, and to revert back to previous settings after
- Enhances tests for the above refactor

## TODO

## Issues addressed
Fixes https://github.com/learningequality/morango/issues/170

### DIFF
--- a/morango/errors.py
+++ b/morango/errors.py
@@ -76,3 +76,7 @@ class MorangoInvalidFSICPartition(MorangoError):
 
 class MorangoSkipOperation(MorangoError):
     pass
+
+
+class MorangoDatabaseError(MorangoError):
+    pass
--- a/morango/models/core.py
+++ b/morango/models/core.py
@@ -122,7 +122,7 @@ class InstanceIDModel(models.Model):
     hostname = models.TextField()
     sysversion = models.TextField()
     node_id = models.CharField(max_length=20, blank=True)
-    database = models.ForeignKey(DatabaseIDModel)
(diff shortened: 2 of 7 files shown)


rtibbles (approved):
Code changes make sense - switch to context manager seems to be made in every place that `_set_transaction_repeatable_read` is referenced.

Having the explicit test that runs the 'real' (rather than test) codepath also gives considerable comfort that this works as intended!

PR #171, opened 2022-08-19
Remove deprecation warnings for field attributes and on_cascade.

rtibbles:
## Summary
* Removes reference to deprecated `rel.to` on Foreign Key fields
* Adds `on_cascade` property to all ForeignKey fields.

Existing tests should provide coverage of these changes.

### DIFF
--- a/morango/registry.py
+++ b/morango/registry.py
@@ -19,7 +19,7 @@ from morango.utils import SETTINGS
 
 def _get_foreign_key_classes(m):
     return set(
-        [field.rel.to for field in m._meta.fields if isinstance(field, ForeignKey)]
+        [field.related_model for field in m._meta.fields if isinstance(field, ForeignKey)]
     )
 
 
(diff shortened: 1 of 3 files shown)


morango/registry.py
```diff
@@ -19,7 +19,7 @@
 
 def _get_foreign_key_classes(m):
     return set(
-        [field.rel.to for field in m._meta.fields if isinstance(field, ForeignKey)]
+        [field.related_model for field in m._meta.fields if isinstance(field, ForeignKey)]
```
jamalex, line 22: Ok, yep, looks like related_model [existed back in 1.10 already](https://docs.djangoproject.com/en/1.10/ref/models/fields/), which is our lowest supported Django version, so should be good!

jamalex (approved):
Thanks!

morango/registry.py
rtibbles: The real clue for me is that it is referenced in the function directly below!
```
def _multiple_self_ref_fk_check(class_model):
    """
    We check whether a class has more than 1 FK reference to itself.
    """
    self_fk = []
    for f in class_model._meta.concrete_fields:
        if f.related_model in self_fk:
            return True
        if f.related_model == class_model:
            self_fk.append(class_model)
    return False
```

PR #175, opened 2023-01-31
Rename instance json fields and add ID fields for convenient querying

bjester:
### Background
The `SyncSession` model tracks instance information in existing text fields that contain JSON
### Changes
- Added  `client_instance_id` and `server_instance_id` UUID fields to `SyncSession` that store the IDs of the instances involved in a sync, which simplifies querying against the table when the instance is important info
- Renamed `client_instance` and `server_instance` fields to `client_instance_json` and `server_instance_json` respectively to avoid confusion that these are Django FK fields

## TODO

## Reviewer guidance

Test coverage is pretty good and existing tests passed successfully 

## References
Supports querying by instance for https://github.com/learningequality/kolibri/issues/10019


tests/testapp/tests/helpers.py
```diff
@@ -476,3 +485,12 @@ def setUp(self):
...
         pass
+
+    @classmethod
+    def tearDownClass(cls):
+        # revert migration back to latest migration
+        executor = MigrationExecutor(connection)
+        executor.loader.build_graph()
+        executor.migrate([cls.latest_migration])
+
+        super(TestMigrationsMixin, cls).tearDownClass()
```
bjester, line 496: @rtibbles This was causing the postgres test failures because it wasn't migrated back to the latest state after the migration tests ran. I can put this change in Kolibri's copy too?

morango/migrations/0022_rename_instance_fields.py
```diff
@@ -0,0 +1,25 @@
...
+    dependencies = [
+        ('morango', '0021_store_partition_index_create'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='syncsession',
+            old_name='client_instance',
+            new_name='client_instance_json',
+        ),
```
jamalex, line 19: May need to make some tweaks in Kolibri for this: https://github.com/learningequality/kolibri/search?q=client_instance (should come up during tests after integration)

tox.ini
```diff
@@ -6,7 +6,7 @@ envlist =
 
 [testenv]
 
-whitelist_externals=
+allowlist_externals=
```
jamalex, line 9: 💯

jamalex (approved):
Looks great! I note that there isn't a backfill data migration for old records, but it's probably safest this way (as long as we ensure we don't make assumptions about it not being null -- perhaps adding a regression test in Kolibri for that).

Might be nice to run a batched manual backfill on KDP later, to make querying there easier.

Issue #193, opened 2023-07-20
Allow bypassing of running process check


bjester:
## Overview 
With the ability to resume syncs, we integrated a mechanism that checks whether the process ID it previously ran within is still running. If the sync is initiated by the CLI, this mechanism makes sense. Although, with automated workflows, like a sync processing through a task worker process, it would run into this check if picked up by the same worker process again

#### Description and outcomes
Add an option to the `resume_sync_session` method which bypasses this check:
https://github.com/learningequality/morango/blob/7e8fae5298d028da25c47fce6723d79a7a0a3de1/morango/sync/syncsession.py#L283-L290

roy-ketan:
@bjester can i work on it, if it is not allotted to anyone

bjester:
Hi @roy-ketan,

Thank you for your interest! This issue is already assigned and has an open PR. Our other projects [Kolibri](https://github.com/learningequality/kolibri) and [Kolibri Studio](https://github.com/learningequality/studio) are great places to look for contributor friendly issues.

bjester:
Closed by https://github.com/learningequality/morango/pull/194

PR #188, opened 2023-07-18
Merge up release-v0.6.x to release-v0.7.x

bjester:
## Summary
- Ports changes made in 0.6.17 release to current 0.7.x release branch

## TODO

- [X] Have tests been written for the new code?
- [X] Has documentation been written/updated?
- [ ] New dependencies (if any) added to requirements file

## Reviewer guidance
Do tests pass? 
Do the merge conflict resolutions make sense?

PR #194, opened 2023-07-25
Add optional setting to ignore an existing process on resuming a sync

bjester:
## Background
Morango stores the process ID on the sync session to prevent accidental resumption of a sync that is still active in another process. On Kolibri, if a sync was started via the UI but then fails, when a user tries to manually resume the sync outside of Kolibri's UI and task processing, the check that ensures the process ID is either the same or inactive can fail because the worker processing tasks is likely still a running and active process.

## Summary
- Adds `ignore_existing_process` keyword arg to `resume_sync_session` allowing the bypass of checking the existing process ID

## TODO

## Reviewer guidance
- The test and code review
- This work has been merged into [this PR](https://github.com/learningequality/morango/pull/188), which will run GH action checks against the updates

## Issues addressed
Fixes https://github.com/learningequality/morango/issues/193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ List of the most important changes for each release.
 ## 0.6.17
 - Added `client-instance-id`, `server-instance-id`, `sync-filter`, `push` and `pull` arguments to `cleanupsyncs` management command
 - Added option for resuming a sync to ignore the `SyncSession`'s existing `process_id`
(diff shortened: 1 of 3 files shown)


jamalex (approved):
Sorry for the delay! This looks great from my side -- I had one initial reaction wondering how this would actually be used via the CLI, then I remembered that part lives in Kolibri, so I'm assuming the option to tie into this will live there.

PR #199, opened 2023-09-11
Fix GH action workflows for py2.7 deprecation

bjester:
## Summary
- Updates Morango's python2 linting to be functional
- Updates workflows to have skip checks in each step, to prevent situations for 'waiting to be queued' checks
- Fixes other related issues

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -113,7 +113,11 @@ jobs:
         strategy:
             max-parallel: 3
             matrix:
-                python-version: [2.7, 3.4, 3.5]
+                # Packages are no longer available for the distribution that comes with the
+                # python:3.4 image, so we can't install the necessary dependencies. Eventually,
+                # this will happen to the others as well.
+                # EOL: [3.4]
+                python-version: [2.7, 3.5]
         container:
             image: python:${{ matrix.python-version }}
         steps:
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -18,67 +18,120 @@ jobs:
     unit_test:
         name: Python unit tests
         needs: pre_job
-        if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
         runs-on: ubuntu-18.04
         strategy:
             max-parallel: 5
             matrix:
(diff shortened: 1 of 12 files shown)


.github/workflows/tox.yml
```diff
@@ -18,67 +18,120 @@ jobs:
     unit_test:
         name: Python unit tests
         needs: pre_job
-        if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
         runs-on: ubuntu-18.04
```
rtibbles: This runner image doesn't exist any more - 20.04 is the oldest (this also means that Python 3.4 and 3.5 probably won't work).

bjester: ah!!! so that's why those actions aren't running? I thought maybe the upstream workflow was taking precedence somehow... but this makes more sense. 

I suppose I could update the python2.7 job to use a matrix with 2.7, 3.4, and 3.5, and we use that to load the corresponding python container image. Just thinking that this could open the door to breaking changes in the 0.6.x line and we already released 0.7.0. Or we could just jump to 0.8.x?

bjester: I actually kinda like that idea, if it works, because then going forward we could keep the matrixed job as new python versions are deprecated, and shuffle them out the door through this job.

rtibbles: Yep, that sounds like a good way to do it!

rtibbles:
Looks like the matrixing worked, but now something is breaking in imports in tests, presumably new things in the stdlib not in py3.4?

bjester:
> Looks like the matrixing worked, but now something is breaking in imports in tests, presumably new things in the stdlib not in py3.4?

Yeah the `test.support` backport just needed to be installed, which left the more [troublesome combination for 3.4 + crypto](https://github.com/learningequality/morango/actions/runs/6152499541/job/16694732858?pr=199#step:4:7). The package repositories are no longer live for the most recent python image, as it's beyond LTS. I guess that's reason enough to drop 3.4? with crypto at the least.

rtibbles:
Yeah, 3.4 is a massive pain, 3.5 is tricky, and 2.7 zombies on!

rtibbles (approved):

Issue #183, opened 2023-06-14
cleanupsyncs - IndexError: list index out of range


bjester:
```
ERROR 2022-08-19 14:46:31,145 kolibri.core.tasks.worker Job 2619b37de22a4aed893176fcc1df2663 raised an exception: Traceback (most recent call last):
  File "c:\python36\lib\site-packages\kolibri\core\tasks\worker.py", line 61, in handle_finished_future
    result = future.result()
  File "c:\python36\lib\concurrent\futures\_base.py", line 425, in result
    return self.__get_result()
  File "c:\python36\lib\concurrent\futures\_base.py", line 384, in __get_result
    raise self._exception
  File "c:\python36\lib\concurrent\futures\thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "c:\python36\lib\site-packages\kolibri\core\tasks\job.py", line 117, in execute_job
    result = func(*args, **kwargs)
  File "c:\python36\lib\site-packages\kolibri\core\public\utils.py", line 456, in soud_sync_cleanup
    call_command("cleanupsyncs", ids=clean_up_ids, expiration=0)
  File "c:\python36\lib\site-packages\kolibri\dist\django\core\management\__init__.py", line 131, in call_command
    return command.execute(*args, **defaults)
  File "c:\python36\lib\site-packages\kolibri\dist\django\core\management\base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "c:\python36\lib\site-packages\kolibri\dist\morango\management\commands\cleanupsyncs.py", line 55, in handle
    transfer_session = transfer_sessions[0]
  File "c:\python36\lib\site-packages\kolibri\dist\django\db\models\query.py", line 289, in __getitem__
    return list(qs)[0]
IndexError: list index out of range
```

bjester:
I believe this was fixed with the referenced PR

PR #205, opened 2023-11-14
Don't close sync session unless also older than cutoff

bjester:
## Summary
In order to support sync resumption and automated clean up of incomplete transfers for Kolibri's learn-only device syncing, the command needs to also filter the sync sessions to only those that have been inactive for at least the `expiration` amount of time
- Fixes issue in iterating over sync sessions
- Adds `last_activity_timestamp` filter to sync sessions before iterating to clean them up

## TODO

## Reviewer guidance
Do the tests make sense

## Issues addressed
Fixes https://github.com/learningequality/morango/issues/183

## Documentation

### DIFF
--- a/morango/management/commands/cleanupsyncs.py
+++ b/morango/management/commands/cleanupsyncs.py
@@ -68,7 +68,8 @@ class Command(BaseCommand):
 
         sync_sessions = SyncSession.objects.filter(active=True)
 
-        # if ids arg was passed, filter down sessions to only those IDs if included by expiration filter
+        # if ids arg was passed, filter down sessions to only those IDs
+        # if included by expiration filter
         if options["ids"]:
             sync_sessions = sync_sessions.filter(id__in=options["ids"])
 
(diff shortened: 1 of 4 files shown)


morango/management/commands/cleanupsyncs.py
```diff
@@ -97,8 +98,7 @@ def handle(self, *args, **options):
         transfer_count = transfer_sessions.count()
 
         # loop over the stale sessions one by one to close them out
-        for i in range(transfer_count):
-            transfer_session = transfer_sessions[0]
```
rtibbles, line 101: Oh, so this and below was only ever affecting the first transfer/sync session!

rtibbles (approved):
Changes make sense to me!

morango/management/commands/cleanupsyncs.py
bjester: No I think this was operating as expected. The sync session one though had the problem.

PR #209, opened 2024-01-29
Drop Python 2.7 support

### LINKED ISSUE #207: Drop Python 2.7 support
Kolibri has dropped Python 2.7 support in its develop branch, so Morango can do the same.

All testing and library support for Python 2.7 can be removed, and any special casing added in the code base for Python 2.7 can be removed.

rtibbles:
## Summary

* Removes Python 2.7 support
* Removes testing and linting related to Python 2.7
* Bumps version to 0.8.0
* Removes use of and dependency on future library

## TODO

## Issues addressed

Fixes #207


bjester (approved):
LGTM. I also ran Kolibri's morango integration tests on it to be sure, but nothing blew up as expected.

PR #214, opened 2024-02-06
Clean up residual Python 2.7 code paths

### LINKED ISSUE #207: Drop Python 2.7 support

rtibbles:
## Summary

* Cleans up code paths that check for Python version less than 3

## Issues addressed

Closes #207


bjester (approved):
LGTM

PR #217, opened 2024-03-05
Upgrade Django to 3.2

### LINKED ISSUE #208: Upgrade to Django 3.2
Once Python 2.7 support has been dropped (https://github.com/learningequality/morango/issues/207), we can upgrade the version of Django that Morango uses to 3.2.

In addition to updating Django, this should also update the versions of associated Django libraries to newer versions that are compatible with Django 3.2.

rtibbles:
## Summary
* Upgrade Django to 3.2
* Upgrade DRF
* Remove use of Django six
* Remove no longer used MorangoMPTT model custom handling
* Update fields to remove context argument
* Upgrade MPTT and create new migrations for PositiveIntegerField etc.
* Use python functools. 
* Don't use schema editor as context (no longer allowed for SQLite)
* Update Django middleware name.
* Set now required output field on function.
* Update how we handle custom serialization and deserialization, following Django's own serialization framework that conditionally renders to string or leaves data types raw - allows specific fields to opt in to string serialization.
* Importantly, this handling now updates serialization _and_ deserialization behaviour, which it did not previously.
* Update the syncsession tests to run the additional server in an isolated process, as databases were no longer being properly set in the LiveServerThread for LiveServerTestCase
* Updates the psycopg2-binary package to the latest compatible version

## TODO

## Issues addressed

Fixes #208


rtibbles:
~Oh, whoops, I left a print statement in here too!~

bjester (approved):
LGTM and all the tests are happy :+1:

PR #221, opened 2024-03-06
More crypto requires

rtibbles:
## Summary

* Removes the redundant specification of cryptography package
* Upgrades M2Crypto to latest (which is still Python 3.6 compatible)

## TODO

### DIFF
--- a/requirements/postgres.txt
+++ b/requirements/postgres.txt
@@ -1,2 +1,2 @@
+-r accelerated.txt
 psycopg2-binary==2.8.6
-cryptography==3.3.2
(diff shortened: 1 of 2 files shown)


requirements/postgres.txt
```diff
@@ -1,2 +1,2 @@
+-r accelerated.txt
 psycopg2-binary==2.8.6
-cryptography==3.3.2
```
bjester, line 2: :ok_hand:

bjester (approved):
Kolibri's certificate integration tests are still happy :heavy_check_mark:

PR #284, opened 2026-01-20
Allow sync operations to modify the sync filter

### LINKED ISSUE #282: Allow sync operations to modify the sync filter
## Current behavior

Currently, a sync context [strictly refuses](https://github.com/learningequality/morango/blob/92278a5adff82787417d6aad75d9c39e1cec8e55/morango/sync/context.py#L95-L96) to be updated with a new sync filter via `.update()` when it has already been configured with one. Also, when a client uses a sync controller for managing a sync, the controller may [pass a 'prepared context' to the sync operations](https://github.com/learningequality/morango/blob/92278a5adff82787417d6aad75d9c39e1cec8e55/morango/sync/controller.py#L246-L254) that may be a different object than used during other parts of the sync operation (local versus remote). For example, if a network operation updates the filter on the context, subsequent local operations will not receive the updated filter because the controller uses a composite filter for parity between local and remote syncs.

## Desired behavior

The context classes should allow the `sync_filter` to be updated, if and only if:
- the filter has changed
- the sync is still within the initialization stage (prior to the update)
- the old filter is still a subset of the new filter

When a context receives an acceptable new `sync_filter`, it should:
- save it on itself, and any child contexts (see `CompositeSessionContext`)
- update the filter on the `TransferSession` and save it

The context classes should have a new method `.join(context)` that can allow contexts that had been previously returned  from `.prepare()` to receive those contexts back:
- for all but the `CompositeSessionContext`, this should be a no-op because `.prepare()` returns itself
- for `CompositeSessionContext`, this should call `.update(sync_filter=prepared_context.filter)` to implement desired functionality

The `SessionController` should call the new method above after it invokes middleware (operations) and receives a result.

## Value add

Allows implementors of Morango, like [Kolibri](https://github.com/learningequality/kolibri/pull/14070), to implement dynamic partition filtering by modifying the context during the initialization stages of the sync.

## Possible tradeoffs

Implementations using this functionality could add complexity to the existing partitioning, filtering, and certificate scope verification.

bjester:
## Summary
- Updates `Filter` to:
  - separate existing templating mechanism into a factory method for clarity wrt to add and copy operations
  - deduplicate partition filters when adding them
  - have utilities for copying, adding, and checking exact existence of a filter 
  - have explicit unit tests, mostly written by Gemini :robot: 
- Updates context classes to:
  - define new `.join()` interface in parity with `.prepare()`
  - refactors existing `CompositeSessionContext.update()` logic into `.join()`
  - allow modification of sync filters under conditions defined in the issue
  - better test coverage over the affected functionality
- Adds ignore for Jetbrains project config files

## Release notes
- Deprecates `Filter` constructor usage utilizing parameters-- use `Filter.from_template` instead

## TODO

- [X] Has documentation been written/updated? no docs mentioned `Filter`

## Reviewer guidance
1) Ran Kolibri's morango integration tests with these changes: `INTEGRATION_TEST=1 pytest -x kolibri/core/auth/test/test_morango_integration.py`

## Issues addressed
Closes https://github.com/learningequality/morango/issues/282


morango/models/certificates.py
```diff
@@ -360,45 +361,88 @@ def get_description(self, params):
...
-        self._filter_string = string.Template(template).safe_substitute(params)
-        self._filter_tuple = tuple(self._filter_string.split()) or ("",)
+    def __init__(self, filter_str, params=None):
+        """
+        :param filter_str: The partition filter string, which may have multiple separated by newlines
+        :type filter_str: str
+        :param params: DEPRECATED: USE Filter.from_template() INSTEAD
+        :type params: dict|str
+        """
+        if params is not None:
```
rtibbles, line 371: Worth adding a  deprecation warning here, or not worth the effort just to warn us? :)

morango/models/certificates.py
```diff
@@ -360,45 +361,88 @@ def get_description(self, params):
...
+        """
+        The Filter's addition operator overload
+        :param other: Filter or None
+        :type other: Filter|None
+        :return: The combined Filter
+        :rtype: Filter
+        """
+        if other is None:
+            return self
+        # create a list of partition filters, deduplicating them between the two filter objects
```
rtibbles, line 437: Almost feels like partitions are more semantically sets rather than lists?

morango/sync/controller.py
```diff
@@ -253,6 +253,9 @@ def _invoke_middleware(self, context, middleware):
             # invoke the middleware with the prepared context
             result = middleware(prepared_context)
 
+            # return the prepared context back
+            context.join(prepared_context)
```
rtibbles, line 257: Ah, I think this is what I was looking for!

morango/sync/context.py
```diff
@@ -402,50 +413,47 @@ def prepare(self):
...
         super(CompositeSessionContext, self).update(**kwargs)
         # update children contexts directly, but exclude stage and stage_status
         self._update_attrs(**kwargs)
         # handle state changes after updating children
         self.update_state(stage=stage, stage_status=stage_status)
 
-        # During the initializing stage, we want to make sure to synchronize the transfer session
-        # object between the composite and children contexts, using whatever child context's
-        # transfer session object that was updated on the context during initialization
-        current_stage = stage or self._stage
```
rtibbles, line 422: So the logic that was here is removed in favour of join, mostly? But with some tweaks? And then join is called...

tests/testapp/tests/sync/test_context.py
```diff
@@ -120,7 +153,7 @@ def test_update__no_overwrite__push(self):
     def test_update__basic(self):
         context = TestSessionContext()
 
-        sync_filter = mock.Mock(spec=Filter)
+        sync_filter = Filter("a")
```
rtibbles, line 156: Always feels like a confidence boosting change to switch out a Mock for a real object in a test.

rtibbles (approved):
Tests are almost exclusively additions, so we have good guarantees of consistent behaviour here. The tests that have been updated have mostly been to use real objects instead of mocks, which makes the outcome of the tests feel more robust!

No blockers, just comments and questions!

morango/models/certificates.py
bjester: I can add that :+1:

morango/models/certificates.py
bjester: Yeah, I guess that would make this simpler to just use a set.

morango/sync/context.py
bjester: Yeah since `join()` is called in nearly the same spot that triggered this code, it made sense to just move this logic to `join()` and add the new functionality there.

morango/models/certificates.py
bjester: Hmm, I forgot sets don't maintain order. While the Filter class does not explicitly try to preserve that, it does make sense to do so. To avoid adding a dep for `OrderedSet`, I will leave this as is for now.

PR #287, opened 2026-02-13
Streaming sync serialization

### LINKED ISSUE #192: Add chunking to serializing models into Store
## Overview 
Serializing huge amounts of data into the `Store` could cause large increases in memory usage. Currently `_serialize_into_store` processes batches by model, but loads all of the model's data into memory before saving the bulk writes to the `Store` model. 

#### Deliverables
- Relocate and split up `_serialize_into_store` into smaller functions
- Have the serialization process the model objects in chunks, or via a queryset `iterator()` ([ref](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#iterator))

bjester:
## Summary
- This is step one in a complete revitalization of the sync pipeline
- Adds new stream utilities for managing the processing of sync data in a streaming fashion
  - I looked at several external libraries. Finding a combo that was simple, but still supported python 3.6, was a real challenge. The closest I found was `streamz`, which I unfortunately opted against because it uses tornado
- Refactors `_serialize_into_store` logic into individual classes built upon foundational stream utilities-- so much better for unit testing!
- Reorganizes some dependent code into locations for shared access and no circular references
- Adds `typing-extensions` for backported future typing features
- Updates `MorangoProfileController` to use `sync_filter` kwarg instead of `filter`-- always bothered me it shadowed the built-in
- Adds unit tests for new stream utilities and converted serialization code-- the serialization process as a whole has pretty good coverage
- Replaces usage of `_serialize_into_store` with new `serialize_into_store` streaming replacement
- The new approach does not use `bulk_update` as Django was observed to spend excessive time with it

### Improvements
The changes were evaluated by installing the local version into Kolibri. A dedicated command was created within Kolibri to run solely the serialization step, and then the performance of that command was benchmarked.

Further investigation will be required to determine how to reduce the increased duration.

#### Case 1: existing large dataset
Kolibri was launched with a pre-existing database containing data for about 18,000 users.

| Version: # users | Memory Graph | Peak Mem | Duration |
| --- | --- | --- | --- |
| Before: 18k | Screenshot from 2026-02-25 14-09-38 | 325.7 MB | 12.49 sec |
| After: 18k | Screenshot from 2026-02-25 14-09-04  | 93.5 MB | 39.50 sec  |

#### Case 2: artificial 500 users
Kolibri's `generateuserdata` command was used to generate data for 500 users, which is the maximum the command currently supports.

| Version: # users | Memory Graph | Peak Mem | Duration |
| --- | --- | --- | --- |
| Before:  500 | image | 217.9MB | 5.99 sec |
| After: 500 | image | 87.5 MB | 12.21 sec |

#### Case 3: large dataset reduced -- 1000 users
Since the `generateuserdata` command currently can only generate up to 500 users, the existing large dataset was trimmed down to 1000 users. After manually deleting the other users, `kolibri manage` was executed (no-op) to trigger Kolibri's FK integrity check which deletes the broken records. Note, this probably takes longer due to the deletions, which provides additional insights into the process, even though the deletion processing has not really changed.

| Version: # users | Memory Graph | Peak Mem | Duration |
| --- | --- | --- | --- |
| Before:  1000 | image| 308.1 MB | 44.54 sec |
| After: 1000 | image | 87.9 MB | 23.63 sec |

#### Case 4: large dataset reduced -- 5000 users
Again, the existing large dataset was trimmed down, this time to 5000 users. Same situation with regards to deletion behavior as in Case (3)

| Version: # users | Memory Graph | Peak Mem | Duration |
| --- | --- | --- | --- |
| Before:  5000 | image| 339.7 MB | 55.88 sec |
| After: 5000 | image| 92.3 MB | 28.92 sec |

## How AI was used
- To look for stream libraries
- Multiple models/providers were used to prototype the stream utilities
- To verify and correct type hinting
- To add comments, edited afterwards
- To create tests for streaming utilities (simplistic)
- To bootstrap tests for the serialization stream utils, heavily refactored by me
- To generate documentation

## TODO

## Reviewer guidance
- Install the branch locally to Kolibri and perform some syncs with another local Kolibri

## Issues addressed

Closes https://github.com/learningequality/morango/issues/192

### DIFF
--- a/docs/architecture/index.rst
+++ b/docs/architecture/index.rst
@@ -132,6 +132,48 @@ In Kolibri, on the ``FacilityDataset`` model, we generate the certificate as a f
 There's flexibility in the application layer for determining the validity of a root certificate, and it's specified on a per-profile basis. For the ``facilitydata`` profile, Kolibri leverages its ``auth`` models for this.
 
 
+Streaming architecture
+----------------------
+
+Morango includes a streaming architecture for memory-efficient processing of sync data. This architecture is implemented in the ``morango.sync.stream`` module and provides a modular, ETL-like pipeline pattern for processing data records one-by-one, significantly reducing memory overhead compared to batch processing approaches.
+
+The streaming architecture is built around several core concepts:
+
+**Stream modules**
+  Abstract base classes that form the foundation of the streaming pipeline:
+
+  - ``Source``: The starting point of a pipeline that yields data items
+  - ``OperatorModule``: Transform-like modules that process data items
+  - ``Sink``: Terminal modules that consume data items without yielding further output
+  - ``PipelineModule``: Modules that can be connected to other modules via the ``pipe()`` method
+
+**Pipeline composition**
+  Modules are connected using a fluent interface via the ``pipe()`` method, creating a directed flow of data:
+
+  .. code-block:: python
+
+      source.pipe(transform1).pipe(transform2).end(sink)
+
+**Key pipeline modules**
+  Several specialized pipeline modules are provided:
+
+  - ``Transform``: Applies a 1:1 transformation to each item
+  - ``FlatMap``: Maps each item to zero or more output items
+  - ``Buffer``: Collects items into fixed-size chunks for batch operations
+  - ``Unbuffer``: Flattens chunks back into individual items
+
+**Serialization pipeline**
+  The serialization process uses this streaming architecture through the ``serialize_into_store()`` function, which constructs a pipeline that:
+
+  1. Reads dirty app models from the database (``AppModelSource``)
+  2. Buffers records for efficient database lookups (``Buffer``)
+  3. Looks up corresponding store records (``StoreLookup``)
+  4. Updates store records with new data (``StoreUpdate``)
+  5. Buffers by model type for efficient bulk operations (``ModelPartitionBuffer``)
+  6. Writes changes to the database (``WriteSink``)
+
+This streaming approach ensures that memory usage remains constant regardless of dataset size, making Morango suitable for large-scale deployments with limited resources.
+
 Session controller, contexts, and operations
 --------------------------------------------
 
@@ -142,4 +184,3 @@ A unidirectional sync has several stages: ``INITIALIZING``, ``SERIALIZING``, ``Q
 .. image:: ./session-controller-seq.png
 
 The list of operations for each stage are configured through Django settings. The configuration key for each stage follows the pattern ``MORANGO_%STAGE%_OPERATIONS``, so the list/tuple of operations for the ``QUEUING`` stage access the ``MORANGO_QUEUING_OPERATIONS`` configuration value. Built-in operations implement a callable ``BaseOperation`` class by overriding a ``handle`` method. The ``BaseOperation`` class supports raising an ``AssertionError`` to defer responsibility to the next operation.
-
--- a/morango/sync/stream/core.py
+++ b/morango/sync/stream/core.py
@@ -1,0 +1,196 @@
+"""
+Foundational classes for streaming ETL-like pipelines.
+
+Provides a modular source > transform > sink pattern where models are streamed one-by-one through
+a pipeline of connected modules, reducing memory overhead.
+"""
+import abc
+from typing import Any
+from typing import Generic
+from typing import Iterable
+from typing import Iterator
+from typing import List
+from typing import Optional
+from typing import TypeVar
+
+T = TypeVar("T")
+
+
+class StreamModule(abc.ABC):
+    """
+    Abstract base class for all stream modules
+    """
+
+    pass
+
+
+class PipelineModule(StreamModule):
+    """
+    Abstract base class for all stream modules that can be read,
+    or rather can pipe to another module.
+    """
+
+    @abc.abstractmethod
+    def pipe(self, other: "StreamModule") -> "StreamModule":
+        """Connect this module to another module, returning a Pipeline."""
+        pass
+
+
+class OperatorModule(StreamModule):
+    """
+    Abstract base class for all pipeline transform-like modules.
+
+    Each module receives an iterable of items and yields transformed items. Modules are composable
+    using a `Pipeline`, allowing them to be chained together via its pipe method.
+    """
+
+    @abc.abstractmethod
+    def __call__(self, items: Iterable[Any]) -> Iterator:
+        """Process the incoming iterable and yield output items."""
+        pass
+
+
+class Source(PipelineModule, Generic[T]):
+    """
+    A module that represents a streaming pipeline source.
+    """
+
+    def begin(self) -> None:
+        """Called once before the stream begins."""
+        pass
+
+    @abc.abstractmethod
+    def stream(self) -> Iterator[T]:
+        """The primary read method that returns the iterator stream of items."""
+        pass
+
+    def pipe(self, other: "OperatorModule") -> "Pipeline":
+        """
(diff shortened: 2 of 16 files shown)


morango/sync/stream/core.py
```diff
@@ -0,0 +1,196 @@
...
+    Abstract base class for all pipeline transform-like modules.
+
+    Each module receives an iterable of items and yields transformed items.
+    Modules are composable: they can be chained together via the pipe method.
+    """
+
+    @abc.abstractmethod
+    def __call__(self, items: Iterable[Any]) -> Iterator:
+        """Process the incoming iterable and yield output items."""
+        raise NotImplementedError
```
rtibbles: Small inconsistency here and below where raise NotImplementedError is used? Fairly sure it's not needed in addition to the abstractmethod decorator?

docs/architecture/index.rst
```diff
@@ -132,6 +132,48 @@ In Kolibri, on the ``FacilityDataset`` model, we generate the certificate as a f
...
+  - ``PipelineModule``: Transform-like modules that process data items
+  - ``Sink``: Terminal modules that consume data items without yielding further output
+  - ``ReaderModule``: Modules that can be connected to other modules via the ``pipe()`` method
+
+**Pipeline composition**
+  Modules are connected using a fluent interface via the ``pipe()`` method, creating a directed flow of data:
+
+  .. code-block:: python
+
+      source.pipe(transform1).pipe(transform2).end(sink)
```
rtibbles, line 155: Not at all necessary, but the thought of being able to construct the pipeline with pipe operators amused me!
```
source | transform1 | transform2 | sink
```

morango/sync/stream/core.py
```diff
@@ -0,0 +1,196 @@
...
+
+class StreamModule(abc.ABC):
+    """
+    Abstract base class for all stream modules
+    """
+
+    pass
+
+
+class ReaderModule(abc.ABC):
```
rtibbles: Is there a reason this isn't a subclass of `StreamModule`?

morango/sync/stream/core.py
```diff
@@ -0,0 +1,196 @@
...
+    or rather can pipe to another module.
+    """
+
+    @abc.abstractmethod
+    def pipe(self, other: "StreamModule") -> "StreamModule":
+        """Connect this module to another module, returning a Pipeline."""
+        pass
+
+
+class PipelineModule(StreamModule):
```
rtibbles: Maybe `TransformModule`? Or seeing that you use that more specifically below `OperatorModule`?

morango/sync/stream/core.py
```diff
@@ -0,0 +1,196 @@
...
+    def consume(self, item: T) -> None:
+        """Process the incoming item from the stream."""
+        raise NotImplementedError
+
+    def finalize(self) -> None:
+        """Called once after all items have been consumed."""
+        pass
+
+
+class Pipeline(ReaderModule):
```
rtibbles: I think the use of Pipeline here and `PipelineModule` above does make it feel more confusing to me.

morango/sync/stream/serialize.py
```diff
@@ -0,0 +1,408 @@
...
+                stores_to_create, ignore_conflicts=True
+            )
+            for created_store in created_stores:
+                # if bulk_create has not marked it as saving been added, then it must have been
+                # a conflict, so we'll add it to the update list
+                if created_store._state.adding:
+                    stores_to_update.append(created_store)
+
+        if stores_to_update:
+            # TODO: bulk_update performs poorly-- is there a better way?
```
rtibbles, line 284: This library claims 8x speed up over bulk_update - but also doesn't seem to be hugely well maintained, so might be useful to look at for inspiration rather than usage! https://github.com/netzkolchose/django-fast-update

setup.py
```diff
@@ -29,6 +29,7 @@
         "djangorestframework>3.10",
         "django-ipware==4.0.2",
         "requests",
+        "typing-extensions==4.1.1",
```
rtibbles, line 32: I assume this was purposeful, but flagging that this is precisely the same version of typing-extensions that Kolibri bundles (although it's still not quite clear to me what requires it, as it's not a direct dependency).

rtibbles (commented):
Implementation makes sense to me, and I can follow the mapping from existing operation code to the new stream architecture. The minimal changes to the existing operations tests give confidence against regressions.

The only thing I got hung up on was the names of the abstract base classes!

morango/sync/stream/core.py
bjester: Yeah this shouldn't need to raise an error.

docs/architecture/index.rst
bjester: Yes the AI solutions allowed for this, but since those just utilized an underlying `.pipe` method, it was just an alternative and unused API.

morango/sync/stream/serialize.py
bjester: Thanks! Yeah something like updating from a temp table should work better and bypass a lot of `CASE` handling Django does with `bulk_update`

setup.py
bjester: It's used for the `Literal` typing here. This version is the latest to support 3.6. It's possible this requirement definition could be improved so it ensures we're using the latest version for a version of python instead?

morango/sync/stream/core.py
bjester: I think I missed updating the inherited class here after some refactoring. I believe it should be yeah.

setup.py
rtibbles: Yes - although testing it against the version it will be bundled with in Kolibri does have its merits, so I think this is fine!

rtibbles (approved):
The naming conventions for the base classes make sense to me, and the benchmarks speak for themselves!

Ran the Kolibri morango integration tests with this installed locally, and everything passed, so I feel good about merging this.

morango/sync/stream/serialize.py
jerch: > ... but also doesn't seem to be hugely well maintained ...

Care to explain why you think it is not well maintained? (disclaimer - I am the maintainer :rofl: )

Wouldn't it be better to improve it instead of working on just another manual solution? Of course I don't know your usage constraints and whether it would fit the needs.

rtibbles: > Care to explain why you think it is not well maintained?

It hasn't been updated for 7 months, doesn't support Python 3.14, and is still in beta with TODOs - this is what drove my assessment.

I certainly didn't mean it as a critique, I am sure it is not top of your priority list to make this universally usable - we have maintained packages that are used more broadly than in our work, and making the time to make it usable for every external use case has usually been outside of our capacity as well.

Thanks for making the package, and if we can upstream anything, we'll be sure to do so!

jerch: Oh well, seems I need to be more aggressive in advertising ready version numbers so ppl and AIs think it is stable.
The beta state is simply still there because of certain features, where I hoped for feedback to layout the directions. The core functionality (fast_update impl) has been stable and actually TDD-tested since alpha versions (several years ago). And you are right - I don't test for new python versions, as long as it is an incremental update and the package doesn't use C-API. Django LTS versions get tested though.

And 7 months without version development does not mean it is abandoned. Can also mean - it just works as intended :wink: 

(There is actually a psycopg3 COPY rewrite in the making, but I got stuck on a psycopg issue.)

bjester: I will definitely take a look at it!

PR #296, opened 2026-03-19
Migrate morango to uv for package management

bjester:
## Summary
- Converts projet to `project.toml` format
  - Consolidates requirements into `project.toml`
  - Sets it up to use `setuptools-scm` which uses repository tags for versioning 
  - The build should look at the most recent tags for the parent, hopefully to ensure we can release from different release branches
- Migrates the project to `uv` for dependency management and builds
  - Updates all testing workflows to pre-build the wheel using python 3.9, which mimics our build process
  - EOL python workflows continue to use python image, and all others use `uv` to set up python
  - Uses `uv build` to generate the distribution files
  - Uses `--installpkg` with pre-built wheel for `tox` to avoid having it build Morango itself, and the issues that causes with EOL versions of python
- Updates the Read the Docs build to use `uv`
- Migrates the project to use `prek` instead of `pre-commit`
- Adds some README documentation about getting set up for development

## TODO

## Reviewer guidance
- We'll have to test the release process manually, unless there's a different suggestion.
- I verified the built wheel file does not contain any unnecessary files not present previously

## AI usage
Local AI was used to do the initial migration. Claude was used to verify the approach, and make some corrections. Finally, manual trial and error were needed to fix what AI could not.


.github/workflows/check_migrations_sqlite.yml
```diff
@@ -12,10 +12,34 @@ jobs:
         uses: fkirc/skip-duplicate-actions@master
         with:
           github_token: ${{ github.token }}
-          paths: '["morango/migrations/*.py", ".github/workflows/check_migrations_sqlite.yml", "setup.py", "requirements/*.txt"]'
+          paths: '["morango/migrations/*.py", ".github/workflows/check_migrations_sqlite.yml", "pyproject.toml"]'
+  build:
```
marcellamaki, line 16: just want to see if i'm understanding this well from the reviewer guidance. so this part is new because it gives us more similarity (and.... therefore increased confidence?) to the build process for morango releases?

.github/workflows/tox.yml
```diff
@@ -12,39 +12,63 @@ jobs:
...
+    needs: [pre_job, build]
     runs-on: ubuntu-latest
     strategy:
       max-parallel: 5
       matrix:
         python-version: [3.9, '3.10', '3.11', '3.12', '3.13', '3.14']
     steps:
       - uses: actions/checkout@v6
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Download wheel artifact
```
marcellamaki, line 50: and here, it uses that whl to help with the python version management that you were describing

README.md
```diff
@@ -14,14 +14,36 @@ Developed in support of the [Kolibri](https://github.com/learningequality/kolibr
...
 ## Developer documentation
 
-See [morango.readthedocs.io](https://morango.readthedocs.io)
+See [morango.readthedocs.io](https://morango.readthedocs.io) for documentation on how Morango works.
+
+### Getting started
+
+To start contributing to Morango, first make sure you [have `uv` installed](https://docs.astral.sh/uv/getting-started/installation/).
+
+Create a virtual environment, with at least python 3.9:
```
marcellamaki, line 23: 3.10?

pyproject.toml
```diff
@@ -0,0 +1,148 @@
...
+
+[project]
+name = "morango"
+dynamic = ["version"]
+description = "Pure Python sqlite-based Django DB replication engine."
+authors = [{name = "Learning Equality", email = "dev@learningequality.org"}]
+maintainers = [{name = "Learning Equality", email = "dev@learningequality.org"}]
+readme = "README.md"
+license-files = ["LICENSE"]
+requires-python = ">=3.6,<3.15"
```
marcellamaki, line 13: so we are still supporting those versions, we are just using 3.10 as the default now?

marcellamaki (commented):
I have added some questions, which I think reflect more about my non-understanding of morango and python version management than anything about uv in particular 😂 thank you for bearing with be @bjester

.github/workflows/check_migrations_sqlite.yml
bjester: Yeah this is the result of the complexity in that with older versions of python, we can't simply use `pip install .` for Morango having the more modern `pyproject.toml`. It needs more modern build tools that don't support older versions of python. 

Morango's build and publish _always_ uses python 3.9, so this mimics that to create a wheel file which CAN be installed later (line 88):
```
pip install $(ls dist/*.whl | head -n 1)[test]
```
- `$(ls dist/*.whl | head -n 1)` this is just a subshell to use globbing since `pip` wants a full file name
- `[test]` this tells pip to install the `test` extras defined in the `pyproject.toml`

So the result is that this is indeed more realistic tests by testing the wheel that we'd publish (built on 3.9) instead of building a wheel within each python version being tested.

.github/workflows/tox.yml
bjester: Yeah this copies that wheel file artifact to make it available to this step.

pyproject.toml
bjester: The build tools require at least 3.9, which is what we used to build and publish the wheel before these changes. So yes the installation still supports these versions, just that development needs newer versions.

README.md
bjester: Yeah is this confusing? Perhaps it should just use 3.9. This is the minimum for build tools

.github/workflows/check_migrations_sqlite.yml
rtibbles: Interesting - this might make doing a parallel version in Kolibri slightly more challenging, because there we build the whl file in Python 3.6 to bundle the dependencies (I am sure we'll find a work around, but this is good extra context).

README.md
marcellamaki: I think the "at least 3.9" and then having it use 3.10 is what confused me. if using 3.10 is better, maybe just adding a little bit of extra detail to the comment about why that's preferable (but confirming that 3.9 still works)? that said, i definitely am not someone who has a lot of context here, so this might be much more apparent to you and others (prathamesh, richard) who are working with morango more consistently. it's definitely not a blocker, just a thought

pyproject.toml
marcellamaki: got it - thanks!

.github/workflows/check_migrations_sqlite.yml
```diff
@@ -12,10 +12,34 @@ jobs:
...
+        if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: wheel
+          path: dist/*.whl
+          retention-days: 1
   migration_test:
     name: SQLite migration tests
-    needs: pre_job
+    needs: [pre_job, build]
```
rtibbles, line 42: I don't think it's a concern in morango - but once concern I've had in Kolibri is that our workflows are overly granular.

I've experimented with having more top level "CI" in some repos recently - so that our CI checks become more of a cascade (make sure it builds first, then test kind of thing).

With no good way in Github Actions to group workflows better, I am tempted to do more of this - but morango doesn't have enough for this to be a concern!

pyproject.toml
```diff
@@ -0,0 +1,148 @@
+[build-system]
+requires = ["setuptools>=64", "setuptools-scm>=8"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "morango"
+dynamic = ["version"]
+description = "Pure Python sqlite-based Django DB replication engine."
```
rtibbles, line 8: Is it SQLite based? :)

(I know this is copied, but just funny given how much work we've done to make sure it works on postgres too!)

rtibbles (approved):
Code changes make sense - I did some spot checks on the different CI jobs, and all the tests seemed to be running with the correct packages installed and on the right Python versions, so all the changes seemed to have successfully maintained what is required!

README.md
bjester: Yeah I think I should update to say 3.10. Python 3.9 is technically EOL already!

pyproject.toml
bjester: Haha good catch

PR #313, opened 2026-04-02
Fix silent deserialization failure with unique together constraint

### LINKED ISSUE #306: Conflict from `unique_together` expression silently fails deserialization
**Target branch:** release-0.8.x

## Observed behavior

Observed on Kolibri, written up with some simplifications.

### Data model
A syncable Django model `User` has a unique constraint on two fields:
- `dataset_id`
- `picture_password`

__Ignore security concerns. This has a very purposeful use case in Kolibri.__

### Scenario
1. Two devices, A and B, both create individual `User` records on their devices for the same `dataset_id` and `picture_password`. 
2. Devices A and B use Morango to sync with each other
3. Since two `User` records have the constraint to prevent duplicate `picture_password` for the same `dataset_id`, the records should fail to deserialize during the sync
4. After the sync completes, Device A does not have the conflicting `User` record for the user created on Device B, and vice versa (expected due to the unique constraint)
5. Reviewing the sync logs, no warnings were shown about failing to deserialize any models
6. Checking the Morango `Store`, the user records do exist, but do not have any `deserialization_error` and also do not have `dirty_bit=True` (😱 )

## Errors and logs

None observed!

## Expected behavior

1. Morango logs warnings about the `Store` records that failed to deserialize to their app models (`User`)
2. Morango adds a `deserialization_error` to the store records annotating the error
3. This scenario should not fail the entire sync

## User-facing consequences

The silent failure to deserialize makes it appear that data was lost, which would be a cause for concern among end-users.

## Steps to reproduce

Suggestion is to write a regression test and prove this scenario, then address fixing it.

## Context

Morango 0.8
Kolibri 0.19

ozer550:
## Summary
- Added IntegrityError to the caught exceptions in the deserialization pipeline so unique constraint violations are handled instead of silently ignored.
- Replaced the bulk upsert  in the non-self-referential branch with per-record saves.
- Added regression tests in test_controller.py for the unique constraint scenario.

## TODO

## Reviewer guidance
- Check if the regression tests make sense!
- Manually syncing two devices with voilating unique constraint for fields.

## Issues addressed
closes #306


morango/sync/operations.py
```diff
@@ -228,14 +432,54 @@ def _validate_store_foreign_keys(from_model_name, fk_references):
...
+    Attempt to save one deserialized app model into the app table.
+
+    On success: clears dirty_bit and deserialization_error on the Store record.
+    On failure: records the error on the Store record, keeps dirty_bit True,
+    and logs a warning.
+
+    :returns: True if save succeeded, False otherwise
+    """
+    try:
+        with transaction.atomic():
```
ozer550, line 447: This is so that the outer transaction stays healthy, and we can keep saving other records.

morango/sync/operations.py
```diff
@@ -284,24 +530,25 @@ def _deserialize_from_store(profile, skip_erroring=False, filter=None):
...
-                            store_model.dirty_bit = False
-                            store_model.deserialization_error = ""
-                            store_model.save(
-                                update_fields=["dirty_bit", "deserialization_error"]
-                            )
                         except (
                             exceptions.ValidationError,
                             exceptions.ObjectDoesNotExist,
                             ValueError,
                         ) as e:
```
bjester, line 537: Perhaps we don't need this `try`/`except` anymore? So we can simplify this, leveraging the behavior of `_save_deserialized_record`

morango/sync/operations.py
```diff
@@ -284,24 +530,25 @@ def _deserialize_from_store(profile, skip_erroring=False, filter=None):
...
+                        if app_model:
+                            _save_deserialized_record(
+                                store_model, app_model, model.__name__, excluded_list
+                            )
+                        else:
+                            store_model.dirty_bit = False
+                            store_model.deserialization_error = ""
+                            store_model.save(
+                                update_fields=["dirty_bit", "deserialization_error"]
+                            )
```
bjester, line 551: So you're right, the `app_model` may be `None`, so this makes sense

morango/sync/operations.py
```diff
@@ -543,7 +760,7 @@ def _queue_into_buffer_v2(transfersession, chunk_size=200):
     We use raw sql queries to place data in the buffer and the record max counter buffer, which matches the conditions of the FSIC.
     """
     sync_filter = Filter(transfersession.filter)
-    with begin_transaction(sync_filter, shared_lock=True):
+    with _begin_transaction(sync_filter, shared_lock=True):
         server_fsic = json.loads(transfersession.server_fsic)
         client_fsic = json.loads(transfersession.client_fsic)
```
bjester, line 749: Please restore comment

morango/sync/operations.py
bjester, line 763: Please restore comment

morango/sync/operations.py
```diff
@@ -370,40 +610,17 @@ def _deserialize_from_store(profile, skip_erroring=False, filter=None):
...
-                        )
-
-                # clear dirty bit for all store records for this model/profile except for rows that did not validate
-                store_models.exclude(id__in=excluded_list).filter(
-                    dirty_bit=True
-                ).update(dirty_bit=False)
+                        continue
+                    _save_deserialized_record(
+                        store_model, app_model, model.__name__, excluded_list
+                    )
```
bjester, line 627: The `excluded_list` doesn't seem necessary here, so maybe it's made optional for this function?

morango/sync/operations.py
```diff
@@ -228,14 +432,56 @@ def _validate_store_foreign_keys(from_model_name, fk_references):
...
+    """
+    Attempt to save one deserialized app model into the app table.
+
+    On success: clears dirty_bit and deserialization_error on the Store record.
+    On failure: records the error on the Store record, keeps dirty_bit True,
+    and logs a warning.
+
+    :returns: True if save succeeded, False otherwise
+    """
+    from django.db import transaction as django_transaction
```
bjester: Strange, I did see the issue the CI. We already have `transaction` imported. Once the other stuff is tackled, it might be worth dropping the commit to see if it fails again.

bjester (changes requested):
This is really mostly complete, just some cleanup and tidying code.

And the small `deserialization_exception` enhancement!

morango/sync/operations.py
ozer550: I initially tried removing the try/except around _deserialize_store_model to reduce code, but I think keeping it is actually better. The reason is that there are two different things that can go wrong at two different moments. the json from the Store might fail having a deserialization failure. Second, our unique constraint violation case. clarity is worth the few extra lines?

morango/sync/operations.py
ozer550: We do execute  `excluded_list.append(store_model.id)` to keep track of shared failed ids?

bjester: Yes but the purpose for that list is this very loop. So modifying the list is a bit redundant, and theoretically makes the `in` check slower every iteration without any benefit.

The `store_model.id` is unique and the same as `app_model.id`, so as this loops through the store+app models, if the ID gets added to the `excluded_list`, it will never come up again in the loop.

bjester (approved):
Looks great! The TDD driven approach worked really well here. Thanks for tackling this @ozer550

Issue #315, opened 2026-04-14
Track deserialization exception type for targeted handling


bjester:
## Overview

**Complexity:** Low
**Target branch:** release-0.8.x

### Context

When a deserialization error occurs, Morango stores the error message on the `Store` record in the `deserialization_error` field.

### The Change

It would be convenient if Morango had another field, perhaps `deserialization_exception`, which provides the fully-qualified exception module path of the error that occurred during deserialization. The `deserialization_error` would continue to be a string message of the error. The new field would allow more programmatic handling of these errors, and avoid parsing the error message.

### Acceptance Criteria

- [ ] A new `deserialization_exception` field should be added to `Store`
- [ ] Migrations should exist for this change
- [ ] The deserialization process should store the full exception module path whenever it currently sets `deserialization_error`
- [ ] Tests should cover this change

### Out of Scope

Handling specific errors

### References

Followup to https://github.com/learningequality/morango/issues/306

bjester:
Closed by https://github.com/learningequality/morango/pull/316

PR #316, opened 2026-04-14
Add deserialization_exception field for programmatic error handling

bjester:
## Summary
This allows programmatic handling of deserialization errors instead of parsing error messages

- Adds a new `deserialization_exception` nullable field to `Store` model that stores the fully-qualified exception class path when deserialization fails
- Updates existing `deserialization_error` field to be nullable
- Updates handling of deserialization errors to set both fields, or otherwise null both of them
- Adds tests for changes
- Removes migration tests that verified the squash migration, which now cause issues with new migrations

## References
Followup to #306
Closes https://github.com/learningequality/morango/issues/315

## AI Usage
Directed an AI agent to tackle the issue, iterated on its work for a loop, then did a small refactor myself. Had the AI agent add test coverage, then reviewed.


morango/sync/operations.py
```diff
@@ -512,7 +517,20 @@ def _deserialize_from_store(profile, skip_erroring=False, filter=None):
 
             # if requested, skip any records that previously errored, to be faster
             if skip_erroring:
-                store_models = store_models.filter(deserialization_error="")
+                # previously, deserialization_error was not nullable, and set to empty string if
+                # there wasn't an error, so this nullifies empty strings to ensure the filter
+                # applies either way
+                store_models = (
```
ozer550, line 523: Doesn't this keep the errored rows instead of skipping them? as mentioned in the comment above?
`# if requested, skip any records that previously errored, to be faster`

ozer550: Can we also add a test to verify this `skip_erroring` behavior after the null string change?

ozer550 (changes requested):
I might be wrong but it seems like we are doing opposite of "skip rows that previously failed deserialization" ? Was this change in behavior intended? Also adding some tests to verify the null string change for skipping would be useful!

morango/sync/operations.py
bjester: Yeah I think you're right, it should be `__isnull=True`. Good catch!

bjester: Fixed! And added tests.

morango/sync/operations.py
ozer550: Doing a grep on `deserialization_error`, would we want the `deserialization_exception` populated here as well?

ozer550 (commented):
I did a grep on places where we had "deserialization_exception" and found a place where we might need to add the execution path as well?

morango/sync/operations.py
bjester: I think you're right, we probably should

bjester:
@ozer550 I've updated it to set the exception in the code you found. I also grepped and found some backend code setting to empty string. We probably didn't to set it there, but I updated the code anyway, to be have parity with `deserialization_error` handling.

morango/sync/operations.py
```diff
@@ -310,33 +316,32 @@ def _validate_missing_store_foreign_keys(from_model_name, to_model_name, temp_ta
...
         c.execute(select_sql)
         for from_field, from_pk, to_pk in c.fetchall():
-            err = dict(
-                {
-                    from_field: "{to_model_name} instance with id '{to_pk}' does not exist".format(
-                        to_model_name=to_model_name,
-                        to_pk=to_pk_field.to_python(to_pk),
-                    )
-                }
+            err = IntegrityError(
```
ozer550, line 330: Is there a reason we are specifically choosing IntegrityError here? This is validation issue rather than DB integrity one? Might lead to misinterpretations? Would it be better to have a special dedicated Morango error class?

ozer550 (commented):
Just a small suggestion not a blocker regarding the Interity Error. Otherwise looks good to me.

morango/sync/operations.py
bjester: Nominally the function "validates", but it validates FKs. The previous implementation was trying to emulate a validation error, but in reality, a missing FK is an integrity issue. Continuing to have it be represented as a validation error, which isn't entirely specific, doesn't benefit the ability to analyze the errors (i.e. new `deserialization_exception`). So this is an improvement by being more specific, which wasn't really necessary before without `deserialization_exception`

ozer550 (approved):
Lgtm! Thanks Blaine.

PR #318, opened 2026-04-15
Streaming deserialize: add source stream and task

### LINKED ISSUE #317: Implement DeserializeTask and StoreModelSource for streaming deserialization
## Current behavior
The `_deserialize_from_store` function in `morango/sync/operations.py` handles deserialization in a monolithic function that mixes multiple concerns. The serialization side has already been migrated to a streaming architecture using `Source > Transform > Sink` modules in `morango/sync/stream/serialize.py`, with `AppModelSource` and `SerializeTask` as foundational components.

## Desired behavior
The deserialization pipeline should have equivalent foundational components: a `DeserializeTask` carrier class that holds context as a store model flows through the pipeline, and a `StoreModelSource` that yields tasks for dirty store models matching a sync filter.

## Deliverables

A new file `morango/sync/stream/deserialize.py` should contain:

**DeserializeTask** carrier class:
- hold a reference to a `Store` model instance
- provide a `model` property that resolves the syncable model class from the store's profile and model_name via `syncable_models`
- track the deserialized app model once processed
- track an FK cache for reuse during deserialization
- track any errors encountered during processing
- provide a `has_errors` boolean property

**StoreModelSource** pipeline source:
- accept the same constructor parameters as `AppModelSource`: `profile`, `sync_filter`, `dirty_only`, and `partition_order`
- iterate through sync filter partitions in the specified order (asc/desc), yielding Q conditions for each prefix
- query `Store` models (not app models) filtered by profile, partition, and dirty_bit
- exclude store records that already have a `deserialization_error` set
- track seen store IDs to avoid yielding duplicates when partitions overlap
- yield `DeserializeTask` objects for each matching store model
- use `.iterator()` for memory-efficient streaming

## Notes

- `StoreModelSource` should mirror the structure of `AppModelSource` in `morango/sync/stream/serialize.py` - same constructor signature, same `prefix_conditions()` pattern, same duplicate-prevention approach
- The source queries `Store` models (the serialized data), whereas `AppModelSource` queries application models - this is the key difference
- Import `Store` from `morango.models.core`, `Source` from `morango.sync.stream.core`, `syncable_models` from `morango.registry`
- Unit tests should be added for both classes
- Tests should verify: partition ordering (asc/desc), dirty_bit filtering, duplicate prevention, deserialization_error exclusion
- This is mostly new code-- the existing deserialization process should not be modified, although since the new source class will share some features with the serialization `AppModelSource`, it could be worthwhile to consolidate some of that into a shared base class

## Value add

Provides the foundational building block for the streaming deserialization pipeline, matching the pattern already established for serialization. Enables modular, testable components instead of a monolithic function.

## Possible tradeoffs

- The `deserialization_error=""` filter is hardcoded - could be made configurable in future if there's a need to retry erroring records
- The `_seen` set grows with the number of store models - should be acceptable for typical sync sizes but worth monitoring for very large datasets

## AI Usage

This issue was created with AI assistance under my guidance. I reviewed and directed all content.

bjester:
## Summary
- Extracts common behavior for streaming source and task classes into new base classes for use with both serialization and deserialization
- Adds new `StoreQuerySet` methods that apply filtering for selecting records based on whether they have deserialization errors. These can be used in Kolibri later
-  Adds method to model registry for querying store records in model dependency order in parity with the method used by serialization
- Creates deserialization stream and task specific classes as foundation for streaming deserialization stage

## TODO

## Reviewer guidance
- Do the tests cover the new and refactored code enough?
- Did I properly translate the self-ref-ordering and deserialization error filtering that were added after this PR was opened in draft? (I rebased it and applied

## Issues addressed

Closes https://github.com/learningequality/morango/issues/317

## AI Usage
I used Claude to do the rebase and bring it up-to-date with the upstream changes. It made a rightful mess of it, and so I rewrote some of the changes to better follow the patterns and then had it clean up the mess by updating tests.

### DIFF
--- a/morango/models/core.py
+++ b/morango/models/core.py
@@ -23,6 +23,7 @@ from django.db.models.deletion import Collector
 from django.db.models.expressions import CombinedExpression
 from django.db.models.fields.related import ForeignKey
 from django.db.models.functions import Cast
+from django.db.models.functions import NullIf
 from django.utils import timezone
 from django.utils.functional import cached_property
 
@@ -440,10 +441,30 @@ class StoreQueryset(models.QuerySet):
             .values_list("fixed_id", flat=True)
         )
 
+    def filter_deserialization_error(self, has_error: bool) -> "StoreQueryset":
+        """
+        Filters the queryset to return Store records that have or have not any
+        deserialization errors
+        """
+        # this nullIf assertion is generally more performant than an OR'd statement on
+        # unindexed columns, in both SQLite and PostgreSQL
+        return self.annotate(
+            _deserialization_error=NullIf(
+                F("deserialization_error"), Value(""), output_field=models.TextField()
+            )
+        ).filter(_deserialization_error__isnull=not has_error)
 
-class StoreManager(models.Manager):
-    def get_queryset(self):
-        return StoreQueryset(self.model, using=self._db)
+    def filter_has_deserialization_error(self) -> "StoreQueryset":
+        """Filters the queryset to return Store records that have deserialization errors"""
+        return self.filter_deserialization_error(True)
+
+    def exclude_has_deserialization_error(self) -> "StoreQueryset":
+        """Filters the queryset to return Store records that have no deserialization error"""
+        return self.filter_deserialization_error(False)
+
+
+class StoreManager(models.Manager.from_queryset(StoreQueryset)):
+    pass
 
 
 class Store(AbstractStore):
--- a/morango/registry.py
+++ b/morango/registry.py
@@ -110,6 +110,18 @@ class SyncableModelRegistry(object):
                 queryset = queryset.order_by(*self._get_nulls_last_ordering(ordering))
             yield queryset
 
+    def get_store_querysets(self, profile) -> Generator[QuerySet, None, None]:
+        """
+        Complementary method to `get_model_querysets` but for Store querysets
+        """
+        from morango.models.core import Store
+
+        for model in self.get_models(profile):
+            store_qs = Store.objects.filter(profile=profile, model_name=model.morango_model_name)
+            if self.get_self_referential_fk(model) is not None:
+                store_qs = store_qs.order_by(*self._get_nulls_last_ordering(("_self_ref_order",)))
+            yield store_qs
+
     @staticmethod
     def _get_nulls_last_ordering(ordering):
         normalized = []
--- a/morango/sync/stream/deserialize.py
+++ b/morango/sync/stream/deserialize.py
@@ -1,0 +1,92 @@
+from typing import Dict
+from typing import Generator
+from typing import List
+from typing import Optional
+from typing import Type
+
+from morango.models.certificates import Filter
+from morango.models.core import Store
+from morango.models.core import SyncableModel
+from morango.registry import syncable_models
+from morango.sync.stream.source import MorangoSource
+from morango.sync.stream.source import SourceTask
+
+
+class DeserializeTask(SourceTask):
+    """Carrier class for providing context through the deserialization pipeline."""
+
+    __slots__ = ("store", "app_model", "fk_cache", "errors")
+
+    def __init__(self, store: Store, fk_cache: Dict):
+        self.store = store
+        self.fk_cache: Dict = fk_cache
+        self.app_model: Optional[SyncableModel] = None
+        self.errors: List[Exception] = []
+
+    @property
+    def id(self) -> str:
+        return self.store.id
+
+    @property
+    def model(self) -> Type[SyncableModel]:
+        return syncable_models.get_model(self.store.profile, self.store.model_name)
+
+    @property
+    def has_errors(self) -> bool:
+        return len(self.errors) > 0
+
+    def set_app_model(self, app_model: Optional[SyncableModel]) -> None:
+        self.app_model = app_model
+
+    def add_error(self, error: Exception) -> None:
+        self.errors.append(error)
+
+
+class StoreModelSource(MorangoSource[DeserializeTask]):
+    """
+    Yields ``DeserializeTask`` objects for dirty store models that match the optional
+    *sync_filter*.
+    """
+
+    def __init__(
+        self,
+        profile: str,
+        sync_filter: Optional[Filter] = None,
+        dirty_only: bool = True,
+        partition_order: str = "asc",
+        fk_cache: Optional[Dict] = None,
+        skip_errored: bool = False,
+    ):
+        """
+        :param profile: The Morango model profile
+        :param sync_filter: The Filter object for this sync
+        :param dirty_only: Whether to filter on dirty records only
+        :param partition_order: Controls how the filter specificity is applied, "asc" or "desc"
+        :param fk_cache: Dictionary cache for FK references
+        :param skip_errored: Whether to skip Store records with deserialization errors
+        """
+        super().__init__(profile, sync_filter, dirty_only, partition_order)
+        self.fk_cache = fk_cache if fk_cache is not None else {}
+        self.skip_errored = skip_errored
+
+    def begin(self) -> None:
+        """Reset fk_cache at the beginning of stream"""
+        super().begin()
+        self.fk_cache.clear()
+
+    def stream_for_filter(
+        self, partition_condition: Optional[str]
+    ) -> Generator[DeserializeTask, None, None]:
+        # the registry yields models in foreign key dependency order, so streaming model by model
+        # ensures a record's foreign key targets are deserialized before it is
+        for store_qs in syncable_models.get_store_querysets(self.profile):
+            qs = store_qs
+            if partition_condition is not None:
+                qs = qs.filter(partition__startswith=partition_condition)
+            if self.dirty_only:
+                qs = qs.filter(dirty_bit=True)
+            if self.skip_errored:
+                qs = qs.exclude_has_deserialization_error()
+
+            for store_model in qs.iterator():
+                yield DeserializeTask(store_model, self.fk_cache)
--- a/morango/sync/stream/source.py
+++ b/morango/sync/stream/source.py
@@ -1,0 +1,104 @@
+import abc
+from typing import Generator
+from typing import Iterator
+from typing import Optional
+from typing import TypeVar
+
+from morango.models.certificates import Filter
+from morango.sync.stream.core import Source
+
+
+class SourceTask(abc.ABC):
+    """Typing for source object passed through streaming pipeline"""
+
+    __slots__ = ()
+
+    @property
+    @abc.abstractmethod
+    def id(self) -> str:
+        pass
+
+
+T = TypeVar("T", bound=SourceTask)
+
+
+class MorangoSource(Source[T], abc.ABC):
+    """
+    Common source functionality for Morango sources, such as SyncableModels and Store records.
+    """
+
+    def __init__(
+        self,
+        profile: str,
+        sync_filter: Optional[Filter] = None,
+        dirty_only: bool = True,
+        partition_order: str = "asc",
+    ):
+        """
+        :param profile: The Morango model profile
+        :param sync_filter: The Filter object for this sync
+        :param dirty_only: Whether to filter on dirty records only
+        :param partition_order: Controls how the filter specificity is applied, "asc" or "desc"
+        """
+        self.profile = profile
+        self.sync_filter = sync_filter
+        self.dirty_only = dirty_only
+        self.partition_order = partition_order
+        self._seen: Optional[set] = None
+
+    def begin(self) -> None:
+        """Initialize seen set at the beginning of the stream"""
+        self._seen = set()
+
+    def prefix_conditions(self) -> Generator[Optional[str], None, None]:
+        """
+        Generates partition prefixes for queries based on the sync filter and partition order.
+
+        This method outputs prefixes in sorted order according to the specified partition
+        order. If no sync filter is provided, it yields `None` to indicate a query
+        without filtering by partition.
+
+        :return: A generator yielding partition prefixes or `None` if no filtering is applied.
+        """
+        if self.sync_filter is None:
+            # yield None once, so we do one query without a partition filter (everything)
+            yield None
+        else:
+            partitions_prefixes = [str(prefix) for prefix in self.sync_filter]
+            partition_iterator = sorted(
+                partitions_prefixes,
+                reverse=self.partition_order == "desc",
+            )
+
+            for prefix in partition_iterator:
+                yield prefix
+
+    def stream(self) -> Generator[T, None, None]:
+        """
+        Streams unique objects based on prefix conditions. This generator method iterates over
+        partition conditions defined in the sync_filter and passes through to `stream_for_filter`
+        to stream back objects, ensuring that only objects with unique `id` values are yielded.
+
+        :return: A generator yielding unique objects.
+        """
+        for partition_condition in self.prefix_conditions():
+            for obj in self.stream_for_filter(partition_condition):
+                # partition filtering could result in overlaps, and since we're walking
+                # through the partitions one by one, we should avoid duplicates. Morango
+                # syncable models and store records have unique IDs across the entire profile
+                if obj.id not in self._seen:
+                    # without sync filters, we do not need to worry about repeating objects
+                    if self.sync_filter is not None:
+                        self._seen.add(obj.id)
+                    yield obj
+
+    @abc.abstractmethod
+    def stream_for_filter(self, partition_condition: Optional[str]) -> Iterator[T]:
+        """
+        This method is intended to generate an iterator that yields data based on the given
+        filtering condition.
+
+        :param partition_condition: A string representing a partition filter prefix condition
+        :return: An iterator yielding items
+        """
+        pass
--- a/tests/testapp/tests/sync/stream/test_deserialize.py
+++ b/tests/testapp/tests/sync/stream/test_deserialize.py
@@ -1,0 +1,217 @@
+import uuid
+
+import mock
+from django.test import SimpleTestCase
+from django.test import TestCase
+
+from morango.models.certificates import Filter
+from morango.models.core import Store
+from morango.models.core import SyncableModel
+from morango.sync.stream.deserialize import DeserializeTask
+from morango.sync.stream.deserialize import StoreModelSource
+
+from ...helpers import StoreFactory
+
+
+class DeserializeTaskTestCase(SimpleTestCase):
+    def setUp(self):
+        self.store = mock.Mock(spec_set=Store)
+        self.store.profile = "test"
+        self.store.model_name = "testmodel"
+        self.task = DeserializeTask(self.store, {})
+
+    @mock.patch("morango.sync.stream.deserialize.syncable_models.get_model")
+    def test_model(self, mock_get_model):
+        model = mock.Mock(spec_set=SyncableModel)
+        mock_get_model.return_value = model
+
+        self.assertEqual(self.task.model, model)
+        mock_get_model.assert_called_once_with("test", "testmodel")
+
+    def test_has_errors(self):
+        self.assertFalse(self.task.has_errors)
+        self.task.add_error(ValueError("bad data"))
+        self.assertTrue(self.task.has_errors)
+
+    def test_set_app_model(self):
+        app_model = mock.Mock(spec_set=SyncableModel)
+        self.task.set_app_model(app_model)
+        self.assertEqual(self.task.app_model, app_model)
+
+
+class StoreModelSourcePrefixConditionsTestCase(SimpleTestCase):
+    """`prefix_conditions` is pure ordering logic over the sync filter and touches no ORM"""
+
+    def test_prefix_conditions__none(self):
+        source = StoreModelSource(profile="test")
+        self.assertEqual(list(source.prefix_conditions()), [None])
+
+    def test_prefix_conditions__with_filter_asc(self):
+        source = StoreModelSource(profile="test", sync_filter=Filter("b\na"), partition_order="asc")
+        self.assertEqual(list(source.prefix_conditions()), ["a", "b"])
+
+    def test_prefix_conditions__with_filter_asc__realistic(self):
+        """
+        Partitions often take the form of `{id}:{additional specificity}`, so the important aspect
+        of partition ordering is that the least specific filter is generally first
+        """
+        source = StoreModelSource(
+            profile="test", sync_filter=Filter("a:test:z\na\na:initial"), partition_order="asc"
+        )
+        self.assertEqual(list(source.prefix_conditions()), ["a", "a:initial", "a:test:z"])
+
+    def test_prefix_conditions__with_filter_desc(self):
+        source = StoreModelSource(
+            profile="test", sync_filter=Filter("a\nb"), partition_order="desc"
+        )
+        self.assertEqual(list(source.prefix_conditions()), ["b", "a"])
+
+
+class StoreModelSourceBeginTestCase(SimpleTestCase):
+    def test_begin_clears_fk_cache(self):
+        """The FK cache is only valid within a single run, since the app models can change"""
+        fk_cache = {"facility": "stale"}
+        source = StoreModelSource(profile="test", fk_cache=fk_cache)
+
+        source.begin()
+
+        self.assertEqual(fk_cache, {})
+        self.assertIs(source.fk_cache, fk_cache)
+
+    def test_begin_initializes_seen(self):
+        """Delegates to the base source, which owns the seen set"""
+        source = StoreModelSource(profile="test")
+        self.assertIsNone(source._seen)
+
+        source.begin()
+
+        self.assertEqual(source._seen, set())
+
+
+class StoreModelSourceStreamTestCase(TestCase):
+    """
+    Streams against real `Store` rows and the real registry, so that the filters are validated as
+    queries the database actually accepts, and not merely as the keyword arguments the source
+    happened to pass along
+    """
+
+    profile = "facilitydata"
+
+    def _store(
+        self,
+        model_name="user",
+        partition="a",
+        dirty_bit=True,
+        deserialization_error=None,
+        self_ref_order=None,
+    ):
+        return StoreFactory(
+            id=uuid.uuid4().hex,
+            profile=self.profile,
+            model_name=model_name,
+            partition=partition,
+            serialized="{}",
+            last_saved_instance=uuid.uuid4().hex,
+            last_saved_counter=1,
+            dirty_bit=dirty_bit,
+            deserialization_error=deserialization_error,
+            _self_ref_order=self_ref_order,
+        )
+
+    def _stream_ids(self, **kwargs):
+        source = StoreModelSource(profile=self.profile, **kwargs)
+        source.begin()
+        return [task.store.id for task in source.stream()]
+
+    def test_stream__dirty_only(self):
+        dirty = self._store(dirty_bit=True)
+        self._store(dirty_bit=False)
+
+        self.assertEqual(self._stream_ids(), [dirty.id])
+
+    def test_stream__dirty_only_false(self):
+        dirty = self._store(dirty_bit=True)
+        clean = self._store(dirty_bit=False)
+
+        self.assertEqual(sorted(self._stream_ids(dirty_only=False)), sorted([dirty.id, clean.id]))
+
+    def test_stream__skip_errored(self):
+        """
+        `deserialization_error` was historically non-nullable and set to an empty string, so both
+        an empty string and null have to be treated as "this record has not errored"
+        """
+        null_error = self._store(deserialization_error=None)
+        empty_error = self._store(deserialization_error="")
+        self._store(deserialization_error="it broke")
+
+        self.assertEqual(
+            sorted(self._stream_ids(skip_errored=True)),
+            sorted([null_error.id, empty_error.id]),
+        )
+
+    def test_stream__errored_included_by_default(self):
+        """Errored records are retried unless the caller opts into skipping them"""
+        errored = self._store(deserialization_error="it broke")
+        clean = self._store(deserialization_error=None)
+
+        self.assertEqual(sorted(self._stream_ids()), sorted([errored.id, clean.id]))
+
+    def test_stream__no_partition_filter(self):
+        first = self._store(partition="a")
+        second = self._store(partition="zzz")
+
+        self.assertEqual(sorted(self._stream_ids()), sorted([first.id, second.id]))
+
+    def test_stream__partition_filter(self):
+        included = self._store(partition="a:one")
+        self._store(partition="b:two")
+
+        self.assertEqual(self._stream_ids(sync_filter=Filter("a")), [included.id])
+
+    def test_stream__deduplicates_across_partition_passes(self):
+        """
+        A less specific partition prefix already matches everything beneath it, so the same record
+        turns up in more than one pass and must only be yielded once
+        """
+        shared = self._store(partition="a:b")
+
+        self.assertEqual(self._stream_ids(sync_filter=Filter("a\na:b")), [shared.id])
+
+    def test_stream__parents_before_children(self):
+        """`facility` is the profile's self-referential model, so its records carry a tree depth"""
+        child = self._store(model_name="facility", self_ref_order=2)
+        root = self._store(model_name="facility", self_ref_order=0)
+        middle = self._store(model_name="facility", self_ref_order=1)
+
+        self.assertEqual(self._stream_ids(), [root.id, middle.id, child.id])
+
+    def test_stream__unresolved_parents_last(self):
+        unresolved = self._store(model_name="facility", self_ref_order=None)
+        root = self._store(model_name="facility", self_ref_order=0)
+
+        self.assertEqual(self._stream_ids(), [root.id, unresolved.id])
+
+    def test_stream__models_in_dependency_order(self):
+        """
+        The registry orders models so that a record's foreign key targets precede it, and the
+        source must preserve that. `user` is registered ahead of `facility` for this profile.
+        """
+        facility = self._store(model_name="facility")
+        user = self._store(model_name="user")
+
+        self.assertEqual(self._stream_ids(), [user.id, facility.id])
+
+    def test_stream__ignores_other_profiles(self):
+        included = self._store()
+        StoreFactory(
+            id=uuid.uuid4().hex,
+            profile="otherprofile",
+            model_name="user",
+            partition="a",
+            serialized="{}",
+            last_saved_instance=uuid.uuid4().hex,
+            last_saved_counter=1,
+            dirty_bit=True,
+        )
+
+        self.assertEqual(self._stream_ids(), [included.id])
--- a/tests/testapp/tests/sync/stream/test_source.py
+++ b/tests/testapp/tests/sync/stream/test_source.py
@@ -1,0 +1,50 @@
+from django.test import SimpleTestCase
+
+from morango.models.certificates import Filter
+from morango.sync.stream.source import MorangoSource
+from morango.sync.stream.source import SourceTask
+
+
+class FakeTask(SourceTask):
+    __slots__ = ("_id",)
+
+    def __init__(self, task_id):
+        self._id = task_id
+
+    @property
+    def id(self):
+        return self._id
+
+
+class FakeSource(MorangoSource[FakeTask]):
+    def __init__(self, *args, task_ids=(), **kwargs):
+        super().__init__(*args, **kwargs)
+        self.task_ids = task_ids
+
+    def stream_for_filter(self, partition_condition):
+        for task_id in self.task_ids:
+            yield FakeTask(task_id)
+
+
+class MorangoSourceBeginTestCase(SimpleTestCase):
+    def test_seen_is_uninitialized_before_begin(self):
+        source = FakeSource("test")
+        self.assertIsNone(source._seen)
+
+    def test_begin_initializes_seen(self):
+        source = FakeSource("test")
+        source.begin()
+        self.assertEqual(source._seen, set())
+
+    def test_begin_resets_seen_between_runs(self):
+        """A source may be streamed more than once, and must not carry state across runs"""
+        source = FakeSource("test", sync_filter=Filter("a"), task_ids=("1", "2"))
+
+        source.begin()
+        first = [task.id for task in source.stream()]
+        self.assertEqual(first, ["1", "2"])
+
+        # without `begin`, every id is already in `_seen` and nothing would be yielded
+        source.begin()
+        second = [task.id for task in source.stream()]
+        self.assertEqual(second, ["1", "2"])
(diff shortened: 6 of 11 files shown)


morango/sync/stream/source.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def stream(self) -> Generator[T, None, None]:
+        """
+        Streams unique objects based on prefix conditions. This generator method iterates over
+        partition conditions defined in the sync_filter and passes thoughts to `stream_for_filter`
+        to stream back objects, ensuring that only objects with unique `id` values are yielded.
+
+        :return: A generator yielding unique objects.
+        """
+        for partition_condition in self.prefix_conditions():
```
rtibblesbot, line 84: **suggestion:** Partition-major iteration drops the model-dependency ordering deserialization relies on.

This loops prefixes outer, models inner, so emission is `prefix1 × (model1..modelN)`, then `prefix2 × (...)`. The comment at `deserialize.py:71-72` claims streaming model by model ensures FK targets come first — that only holds within a single prefix pass.

`_deserialize_from_store` is the opposite shape: models outer (`operations.py:303`) with every prefix OR'd into one `partition__startswith` condition (`operations.py:313-319`), so dependency order is global across the whole filter. Here, a `modelN` record in `prefix1` whose FK target is a `model1` record living only in `prefix2` is emitted before its target.

Ordering is irrelevant for serialization, which is presumably why the base landed this way. For deserialization it is the guarantee. Either invert the loops in `StoreModelSource` (models outer, prefixes OR'd inner, matching the existing code), or drop the comment's claim and document that cross-prefix FK targets are unordered.

morango/sync/stream/source.py
```diff
@@ -0,0 +1,96 @@
...
+from typing import Generator
+from typing import Iterator
+from typing import Optional
+from typing import TypeVar
+
+from morango.models.certificates import Filter
+from morango.sync.stream.core import Source
+
+
+class SourceTask(abc.ABC):
```
rtibblesbot, line 11: **suggestion:** `SourceTask` declares no `__slots__`, so it contributes a `__dict__` to every subclass and `SerializeTask`/`DeserializeTask`'s own slots become inert. `SerializeTask` had working slots before this refactor.

Verified on this class shape:

```
t = T(); hasattr(t, "__dict__")  # True
t.zzz = 5                        # succeeds
```

One task instance exists per record streamed, in a pipeline whose stated point is "reducing memory overhead" (`core.py:1-6`). `abc.ABC` itself sets `__slots__ = ()`; the same line on `SourceTask` restores both the memory saving and the typo-catching.

morango/sync/stream/source.py
```diff
@@ -0,0 +1,96 @@
...
+        :param profile: The Morango model profile
+        :param sync_filter: The Filter object for this sync
+        :param dirty_only: Whether to filter on dirty records only
+        :param partition_order: Controls how the filter specificity is applied, "asc" or "desc"
+        """
+        self.profile = profile
+        self.sync_filter = sync_filter
+        self.dirty_only = dirty_only
+        self.partition_order = partition_order
+        self._seen = set()
```
rtibblesbot, line 51: **suggestion:** `_seen` accumulates one entry per record for the source's lifetime, even when nothing can duplicate. With `sync_filter is None`, `prefix_conditions()` yields exactly one `None`, so `stream_for_filter` runs once and no id can repeat — but a full-store pass still holds every Store id in memory, which is the cost the streaming design exists to avoid.

Skipping the bookkeeping when there is a single prefix condition covers that case. Also note `_seen` is never reset, so a second `stream()` call on the same instance yields nothing.

morango/sync/stream/source.py
```diff
@@ -0,0 +1,96 @@
...
+                reverse=self.partition_order == "desc",
+            )
+
+            for prefix in partition_iterator:
+                yield prefix
+
+    def stream(self) -> Generator[T, None, None]:
+        """
+        Streams unique objects based on prefix conditions. This generator method iterates over
+        partition conditions defined in the sync_filter and passes thoughts to `stream_for_filter`
```
rtibblesbot: **nitpick:** "passes thoughts to `stream_for_filter`" — "through".

morango/sync/stream/deserialize.py
```diff
@@ -0,0 +1,83 @@
...
+
+class DeserializeTask(SourceTask):
+    """Carrier class for providing context through the deserialization pipeline."""
+
+    __slots__ = ("store", "app_model", "fk_cache", "errors")
+
+    def __init__(self, store: Store):
+        self.store = store
+        self.app_model: Optional[SyncableModel] = None
+        self.fk_cache: Dict = {}
```
rtibblesbot: **suggestion:** A fresh `{}` per store record gives the FK cache no reuse. `_deserialize_from_store` creates it once (`operations.py:297`) and threads the same dict through every `_deserialize_store_model` → `cached_clean_fields(fk_cache, ...)` call (`core.py:548`). Per-record, it is always empty on first use, so every FK lookup hits the DB.

Issue #317 asks the task to track an FK cache *for reuse*. A transform could overwrite `task.fk_cache` with a shared dict later, but this default silently defeats the optimization — having `StoreModelSource` own one dict and hand the same reference to every task it yields would match the intent.

morango/sync/stream/deserialize.py
```diff
@@ -0,0 +1,83 @@
...
+    *sync_filter*.
+    """
+
+    def __init__(
+        self,
+        profile: str,
+        sync_filter: Optional[Filter] = None,
+        dirty_only: bool = True,
+        partition_order: str = "asc",
+        skip_errored: bool = True,
```
rtibblesbot: **suggestion:** `skip_errored` defaults to `True`, inverting the legacy default: `_deserialize_from_store(profile, skip_erroring=False, ...)` (`operations.py:283`), plumbed from `SessionController.deserialize_from_store(skip_erroring=False, ...)` (`controller.py:29`).

Many deserialization errors are transient — an `ObjectDoesNotExist` on an FK target arriving in a later sync is the common case — and a record carrying a stale `deserialization_error` is never retried while the flag is on. When this source is wired into the deserialize operation, that quietly changes which records get a second chance.

Either default to `False` to match, or note in the PR body why the flip is intended. Either way the param is missing from the docstring below (which re-documents the four inherited ones) and beyond #317's "same constructor parameters as `AppModelSource`" — worth recording as a deliberate choice.

morango/registry.py
```diff
@@ -110,6 +110,18 @@ def get_model_querysets(self, profile) -> Generator[QuerySet, None, None]:
...
+    def get_store_querysets(self, profile) -> Generator[QuerySet, None, None]:
+        """
+        Complementary method to `get_model_querysets` but for Store querysets
+        """
+        from morango.models.core import Store
+
+        for model in self.get_models(profile):
+            store_qs = Store.objects.filter(
+                profile=profile, model_name=model.morango_model_name
+            ).order_by(*self._get_nulls_last_ordering(("_self_ref_order",)))
```
rtibblesbot: **suggestion:** The `_self_ref_order` sort is applied to every model, including those with no self-referential FK. `get_model_querysets` directly above gates its `order_by` on `morango_ordering` being non-empty; this one doesn't.

`_self_ref_order` is NULL on every row for a model without a self-referential FK (`operations.py:757-759` nulls it out for exactly those), and it is in no index — `Store.Meta.indexes` covers `partition` and `(profile, model_name, partition, dirty_bit)` (`core.py:486-494`). Generated SQL:

```
SELECT ... FROM "morango_store"
WHERE ("model_name" = abc AND "profile" = facilitydata)
ORDER BY "_self_ref_order" ASC NULLS LAST
```

So for most models it forces a full sort of the result set — once per model, per partition prefix — before the first row can be yielded, working against the `.iterator()` streaming this exists for. The registry already caches the answer in `get_self_referential_fk(model)`; gating the `order_by` on it would skip the sort for the majority of models.

morango/models/core.py
```diff
@@ -440,10 +441,30 @@ def char_ids_list(self):
             .values_list("fixed_id", flat=True)
         )
 
+    def filter_deserialization_error(self, has_error: bool) -> "StoreQueryset":
```
rtibblesbot, line 444: **suggestion:** This is a verbatim extraction of `operations.py:326-336`, comment included, but that call site still hand-rolls the annotation. Swapping it for `store_models.exclude_has_deserialization_error()` is one line and keeps the two from drifting — #317's "do not modify the existing deserialization process" is about behaviour, and this is behaviour-preserving.

morango/models/core.py
```diff
@@ -440,10 +441,30 @@ def char_ids_list(self):
...
+        """
+        Filters the queryset to return Store records that have or have not any
+        deserialization errors
+        """
+        # this nullIf assertion is generally more performant than an OR'd statement on
+        # unindexed columns, in both SQLite and PostgreSQL
+        return self.annotate(
+            _deserialization_error=NullIf(
+                F("deserialization_error"), Value(""), output_field=models.BooleanField()
```
rtibblesbot: **nitpick:** `NULLIF(deserialization_error, '')` yields text, not a boolean. It works because only `__isnull` is applied, but the annotation is in the SELECT list too, so any caller reading `_deserialization_error` off a row gets a string from a field declared boolean. Inherited from `operations.py:329-333`; `TextField()` would be accurate in both.

tests/testapp/tests/sync/stream/test_deserialize.py
```diff
@@ -0,0 +1,166 @@
...
+    return qs
+
+
+def _mock_store(store_id):
+    store = mock.Mock(spec_set=Store)
+    store.id = store_id
+    return store
+
+
+class StoreModelSourceTestCase(SimpleTestCase):
```
rtibblesbot: **suggestion:** `StoreModelSourceTestCase` mocks the ORM throughout, so nothing here verifies the filters match real rows. Assertions are on call args (`qs.filter.call_args_list == [mock.call(dirty_bit=True)]`, `qs.filter.assert_not_called()`), which pass for whatever kwargs the source happens to send — `spec=StoreQueryset` catches a missing *method*, not a missing *field*, so a typo'd or renamed field stays green.

`test_stream__preserves_registry_queryset_order` and `test_stream__deduplicates_across_partition_passes` genuinely need the registry stubbed; the filter tests don't. `GetStoreQuerysetsTestCase` in this same PR shows the alternative at about the same length — build a handful of `Store` rows with varying `dirty_bit`/`deserialization_error`/`partition` and assert which ids stream out.

tests/testapp/tests/test_registry.py
```diff
@@ -116,3 +120,89 @@ def test_get_model_querysets_applies_ordering_per_model(self):
         mock_get_ordering.assert_called_once_with(("field_a",))
         queryset_a.order_by.assert_called_once_with("normalized_a")
         queryset_b.order_by.assert_not_called()
+
+
+class GetStoreQuerysetsTestCase(TestCase):
+    """
+    `get_store_querysets` is the deserialization counterpart to `get_model_querysets`, and is
+    responsible for the ordering guarantees the deserialization stage depends upon.
+
```
rtibblesbot, line 130: **praise:** Asserting the rows actually returned rather than generated SQL — with the reasoning written down — is the right level for a guarantee deserialization correctness depends on, and it stays valid across SQLite and Postgres.

rtibblesbot (commented):
PR #318 delivers all four of issue #317's items — `DeserializeTask`, `StoreModelSource`, the shared `MorangoSource` base, and tests for both — plus the `StoreQueryset` error filters and `get_store_querysets`. No blocking defects.

The substantive comments are about semantics the shared base imports into deserialization: partition-major iteration vs. the model-major ordering `_deserialize_from_store` guarantees today, a per-task `fk_cache` that can never hit, a `skip_errored` default that inverts the legacy one, and a lost `__slots__` optimization.

CI: Linting and Migrations pass on `acc181f`; Python tests were still running. The one completed failure (`Handle pull request events` at `Generate App Token`) is infrastructure, unrelated to the diff. Locally `tests/sync/stream/`, `test_registry.py` and `models/test_core.py` pass (106) and `makemigrations --check` is clean.

No UI files, so Phase 3 / manual QA did not apply.

---

</details>

morango/sync/stream/source.py
bjester: There are no cross-partition FKs. The ordering is a naive approach that works with known filters, which prioritize the shortest filters first, which are likely the least specific.

morango/models/core.py
bjester: The code in `operations.py` will eventually be removed. There is no need to update it.

morango/sync/stream/deserialize.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        :param profile: The Morango model profile
+        :param sync_filter: The Filter object for this sync
+        :param dirty_only: Whether to filter on dirty records only
+        :param partition_order: Controls how the filter specificity is applied, "asc" or "desc"
+        :param fk_cache: Dictionary cache for FK references
+        :param skip_errored: Whether to skip Store records with deserialization errors
+        """
+        super().__init__(profile, sync_filter, dirty_only, partition_order)
+        self.fk_cache = fk_cache or {}
```
rtibblesbot: **suggestion:** `{}` is falsy, so `fk_cache or {}` drops a caller-supplied empty dict — and construction is the only time it is passed, so the shared cache I asked for last round never shares. Use `fk_cache if fk_cache is not None else {}`.

No test references `fk_cache`, so reverting to a per-record `{}` stays green. `self.assertIs(tasks[0].fk_cache, cache)` catches both.

morango/sync/stream/source.py
```diff
@@ -0,0 +1,100 @@
...
+        :param profile: The Morango model profile
+        :param sync_filter: The Filter object for this sync
+        :param dirty_only: Whether to filter on dirty records only
+        :param partition_order: Controls how the filter specificity is applied, "asc" or "desc"
+        """
+        self.profile = profile
+        self.sync_filter = sync_filter
+        self.dirty_only = dirty_only
+        self.partition_order = partition_order
+        self._seen = set()
```
rtibblesbot, line 51: **suggestion:** `_seen` and the source-owned `fk_cache` are instance-scoped, but `Pipeline.end()` calls `Source.begin()` once per run (`stream/core.py:112`) and `MorangoSource` does not override it. A reused source with a `sync_filter` yields nothing on a second run; cache entries meaning "FK target validated" never expire, where legacy scopes them to one `_deserialize_from_store` call (`operations.py:297`). Unreachable today, but the new constructor invites callers to hold a source — reset both in `begin()`.

tests/testapp/tests/sync/stream/test_deserialize.py
```diff
@@ -0,0 +1,195 @@
...
+
+        self.assertEqual(self._stream_ids(), [dirty.id])
+
+    def test_stream__dirty_only_false(self):
+        dirty = self._store(dirty_bit=True)
+        clean = self._store(dirty_bit=False)
+
+        self.assertEqual(sorted(self._stream_ids(dirty_only=False)), sorted([dirty.id, clean.id]))
+
+    def test_stream__skip_errored(self):
```
rtibblesbot, line 138: ✅ **Resolved** — addressed in the current code.


**praise:** Both NULL and empty-string `deserialization_error` against real rows — the historical case `NullIf` exists for, and one a mocked queryset could not catch.

rtibblesbot (commented):
PR #318 — all 11 prior findings resolved or acknowledged (verified in the tree, not on thread state); 2 new findings, both on code added last round.

CI: test matrices green; only the contributor-check app-token step fails — infrastructure. No UI files.

`skip_errored` now defaults to `False`, matching `_deserialize_from_store(skip_erroring=False)` rather than #317's text — my request last round, still right.

RESOLVED — morango/sync/stream/source.py:14 — `SourceTask` lacks `__slots__`
RESOLVED — morango/sync/stream/source.py:87 — `_seen` unbounded (never-reset residual refiled below)
RESOLVED — morango/sync/stream/source.py:75 — "passes thoughts to"
RESOLVED — morango/sync/stream/deserialize.py:69 — fresh `{}` per record defeats FK cache
RESOLVED — morango/sync/stream/deserialize.py:58 — `skip_errored` inverted the legacy default
RESOLVED — morango/registry.py:120 — `_self_ref_order` sort applied to every model
RESOLVED — morango/models/core.py:444 — `NULLIF` yields text, not a boolean
RESOLVED — tests/testapp/tests/sync/stream/test_deserialize.py:70 — stream tests mocked the ORM
ACKNOWLEDGED — morango/sync/stream/source.py:80 — partition-major ordering; no cross-partition FKs
ACKNOWLEDGED — morango/models/core.py:444 — `operations.py` duplication; that code is slated for removal
n/a (praise) — tests/testapp/tests/test_registry.py:130 — asserts rows, not SQL

</details>

---

</details>

morango/sync/stream/source.py
```diff
@@ -0,0 +1,100 @@
...
+        order. If no sync filter is provided, it yields `None` to indicate a query
+        without filtering by partition.
+
+        :return: A generator yielding partition prefixes or `None` if no filtering is applied.
+        """
+        if self.sync_filter is None:
+            # yield None once, so we do one query without a partition filter (everything)
+            yield None
+        else:
+            partitions_prefixes = [str(prefix) for prefix in self.sync_filter]
```
rtibblesbot, line 67: **suggestion:** Every prefix in the filter yields a pass, including prefixes already covered by a shorter one — and `StoreModelSource.stream_for_filter` issues one query per registered model per pass.

`Filter("a:test:z\na\na:initial")` (the shape in `test_prefix_conditions__with_filter_asc__realistic`) sorts to `["a", "a:initial", "a:test:z"]`. Pass 1 matches every row the other two can match; passes 2 and 3 are 2 × N-models queries whose rows are all discarded against `_seen`.

Dropping prefixes that extend another prefix in the same filter preserves the covered set exactly — `Filter.contains_partition` (`morango/models/certificates.py:383-385`) already does the test. In the common single-covering-prefix case it also collapses to one pass, which drops the `_seen` bookkeeping and keeps the whole stream in model-dependency order.

Separate from the cross-partition ordering point answered last round; that answer stands.

morango/sync/stream/deserialize.py
rtibblesbot: **suggestion:** (still open) `{}` is falsy, so `fk_cache or {}` silently substitutes a new dict when a caller passes an empty one — the caller then never sees the entries the source populates. `fk_cache if fk_cache is not None else {}` keeps the caller's object.

No test asserts the tasks share the source's dict object; `assertIs(task.fk_cache, source.fk_cache)` would pin both this and the sharing behaviour.

morango/sync/stream/source.py
rtibblesbot, line 51: **suggestion:** (still open) `_seen` — and `fk_cache` in the subclass — are instance state that lives for the source's lifetime, but `Source.begin()` (`morango/sync/stream/core.py:60`) exists so a source can be re-run. `MorangoSource` doesn't override it, so a second `stream()` on the same instance yields nothing: every id is already in `_seen`.

Overriding `begin()` to clear both makes the reuse contract explicit even if the pipeline currently only runs each source once.

rtibblesbot (commented):
PR #318 — 12 of 14 prior findings resolved or acknowledged; 2 still open, plus one new suggestion on `prefix_conditions` (all inline).

HEAD unchanged at `b0b1fd4c` since the last pass. Test jobs pass on 3.6–3.14 across SQLite and Postgres; the only red check is `Handle review requested / Check if author is contributor`, a permissions workflow unrelated to the code. No UI files, so Phase 3 did not apply.

RESOLVED — morango/sync/stream/source.py:14 — `SourceTask` declares no `__slots__`
RESOLVED — morango/sync/stream/source.py:88 — `_seen` accumulates even when nothing can duplicate
RESOLVED — morango/sync/stream/source.py:75 — "passes thoughts to `stream_for_filter`"
ACKNOWLEDGED — morango/sync/stream/source.py:80 — Partition-major iteration drops model-dependency ordering
RESOLVED — morango/sync/stream/deserialize.py:69 — Fresh `{}` per store record gives the FK cache no reuse
RESOLVED — morango/sync/stream/deserialize.py:58 — `skip_errored` inverted the legacy default
RESOLVED — morango/registry.py:121 — `_self_ref_order` sort applied to every model
RESOLVED — morango/models/core.py:453 — `NULLIF` annotation declared `BooleanField`
ACKNOWLEDGED — morango/models/core.py:444 — Legacy call site in operations.py still hand-rolls the annotation
RESOLVED — tests/testapp/tests/sync/stream/test_deserialize.py:70 — `StoreModelSourceTestCase` mocked the ORM throughout
RESOLVED — tests/testapp/tests/sync/stream/test_deserialize.py:116 — Both NULL and empty-string `deserialization_error` against real rows
RESOLVED — tests/testapp/tests/test_registry.py:130 — Asserting returned rows rather than generated SQL
UNADDRESSED — morango/sync/stream/deserialize.py:69 — `fk_cache or {}` drops a caller-supplied empty dict
UNADDRESSED — morango/sync/stream/source.py:47 — `_seen` and `fk_cache` are not reset in `begin()`
</details>

---

</details>

morango/sync/stream/source.py
bjester: That would be a premature optimization that doesn't have a place in this code. What's important to the stream processing is that things are processed in reasonable order. It's up to the implementer to provide a reasonable, reduced, filter set.

bjester (commented):
# Code Review: Streaming deserialize — add source stream and task

## 1. Executive Summary

- **Overall Assessment**: `COMMENT` — no blockers. Two design notes worth resolving before this base class carries the rest of the series.
- **Summary**: All four of #317's deliverables land (`DeserializeTask`, `StoreModelSource`, the shared `MorangoSource`/`SourceTask` base, tests), plus the `StoreQueryset` error filters and `get_store_querysets`. The extraction is genuinely a shared base rather than a copy — `AppModelSource` shrinks from ~45 lines to 12 and keeps behaviour. The test suite for the ordering guarantee runs against real rows on both backends, which is the right level for something deserialization correctness depends on.
- **Impact Surface**: `morango/sync/stream/{source,serialize,deserialize}.py`, `morango/registry.py`, `morango/models/core.py` (`StoreQueryset`/`StoreManager`). No migration needed — no field changes. No existing code path is rewired, so blast radius on this PR alone is limited to the new source; the risk arrives when it is wired into the deserialize operation.
- **Verification**: `pytest tests/testapp/tests/sync/stream/ tests/testapp/tests/test_registry.py tests/testapp/tests/models/test_core.py` → **119 passed**. CI green across 3.6–3.14 / SQLite + Postgres; the only red check is the contributor-check app-token step (infrastructure).

---

## 2. Principle-Grounded Findings

### ⚠️ Architecture & Design (Should-Fix)

- **`stream()` is a partial function — it raises an opaque `TypeError` unless `begin()` was called first** (`morango/sync/stream/source.py:47-51`, `:89`)
  - **Principle**: Defensive Contracts — a public entry point should either work or fail with a message that names the violated precondition.
  - **Issue**: `_seen` is `None` until `begin()` runs, and `stream()` does `obj.id not in self._seen`. Confirmed on this tree:
    ```
    TypeError: argument of type 'NoneType' is not iterable
    ```
    Nothing in that message points at the missing `begin()`. This is also a narrow regression against the pre-refactor `AppModelSource`, which initialised `_seen` in `__init__` and was safe to call standalone — the evidence is in this PR's own diff, where three previously-passing tests had to gain a `source.begin()` line (`tests/.../test_serialize.py:74,118`).
  - **Remediation**: initialise in `__init__` *and* reset in `begin()`. Keeps the reuse contract that `test_begin_resets_seen_between_runs` pins, and makes a stray `stream()` correct rather than fatal:
    ```python
    self._seen: set = set()

    def begin(self) -> None:
        """Reset seen set at the beginning of the stream"""
        self._seen = set()
    ```
    If the `None` sentinel is deliberate — "you must go through `Pipeline.end()`" — then assert it, so the failure names itself: `_assert(self._seen is not None, "begin() must be called before stream()")`.

- **`_seen` grows unbounded on the filtered path — the memory cost this pipeline exists to eliminate** (`morango/sync/stream/source.py:84-93`)
  - **Principle**: the module docstring's own goal — "streamed one-by-one … reducing memory overhead" (`stream/core.py:1-6`).
  - **Issue**: The no-filter case is already exempted, so what remains is exactly the case that matters: a filtered facility sync retains one 32-char id per record streamed, for the lifetime of the source. At ~130 bytes per set entry that is ~130 MB at 1M Store rows — on the low-resource deployments this design targets.
  - **Remediation**: the dedup can be stateless. Passes differ *only* by their partition condition, so a record yielded in pass *k* is precisely a record whose partition matches prefix *k* — excluding earlier prefixes in later passes is exactly equivalent to the id set, and pushes the work into the query:
    ```python
    def stream(self) -> Generator[T, None, None]:
        processed = []
        for prefix in self.prefix_conditions():
            for obj in self.stream_for_filter(prefix, exclude_prefixes=processed):
                yield obj
            processed.append(prefix)
    ```
    with `stream_for_filter` applying `.exclude(partition__startswith=p)` per entry. Prefix counts are small, `processed` is bounded by the filter size rather than the row count, and `_seen` disappears. Holds for `partition_order="desc"` too. This is distinct from the prefix-reduction idea you declined earlier — that one traded query count for coverage reasoning; this one removes state without changing either.

### 💡 Tactical Suggestions & Polish (Optional)

- **`begin()` wipes a caller-supplied `fk_cache`** (`morango/sync/stream/deserialize.py:69-75`): `self.fk_cache.clear()` mutates an object the caller owns, which makes the constructor parameter unable to do the one thing injection is usually for — handing in a warm cache. The remaining value is post-run inspection, which is legitimate but non-obvious. Either say so on the parameter (`:param fk_cache: … cleared at the start of every run`) or only clear when the source created the dict itself.

- **Boolean-flag method sits in the public queryset API** (`morango/models/core.py:444`): `filter_deserialization_error(has_error)` is a flag argument whose two call sites are the intention-revealing wrappers right below it. Renaming it `_filter_deserialization_error` leaves callers with only the two readable spellings. (The wrappers themselves are the right shape — this is just about which of the three is the advertised one.)

- **The "no cross-partition FK references" invariant lives only in a PR thread** (`morango/sync/stream/deserialize.py:80-81`): the comment claims model-major streaming gets FK targets deserialized first, which holds globally only because partitions never contain FKs across each other. That premise is the load-bearing part and it is nowhere in the code. One clause — "partitions do not contain cross-partition FK references, so per-partition model ordering is sufficient" — saves the next reader the rediscovery.

- **`AppModelSource` no longer dedups when `sync_filter is None`** (`morango/sync/stream/source.py:90-91`): a deliberate and near-certainly safe narrowing (a single pass over non-joined querysets cannot repeat a row), but the test that used to cover it was retargeted to the filtered path (`test_serialize.py:92`) rather than replaced, so nothing asserts the unfiltered behaviour any more. Worth a line in the PR body if not a test.

- **Nothing pins `__slots__` effectiveness** (`morango/sync/stream/source.py:11-18`): the `__slots__ = ()` on `SourceTask` is load-bearing and verified working here —
  ```
  DeserializeTask has __dict__: False
  AttributeError: 'DeserializeTask' object has no attribute 'zzz'
  ```
  — but it is one easily-dropped line away from silently reintroducing a `__dict__` on every task in the stream. `with self.assertRaises(AttributeError): task.nope = 1` in `DeserializeTaskTestCase` costs two lines and holds the line for the whole series.

- **`abc.ABC` in `MorangoSource(Source[T], abc.ABC)` is redundant** (`morango/sync/stream/source.py:26`): `Source` → `PipelineModule` → `StreamModule(abc.ABC)` already supplies `ABCMeta`.

- **Docs and CHANGELOG** (`docs/architecture/index.rst:135-176`): the "Streaming architecture" section enumerates the stream module vocabulary and the serialization pipeline; `MorangoSource`, `SourceTask`, and the now-load-bearing `begin()` lifecycle aren't in it. Deferring the prose to the end of the series is reasonable — worth saying so explicitly against the unchecked box rather than leaving it ambiguous.

### ✅ Positive Highlights

- **`GetStoreQuerysetsTestCase` asserts returned rows, not generated SQL** — with the reasoning written into the docstrings, and staying valid across SQLite and Postgres. `test_orders_unresolved_parents_last` in particular pins nulls-last for `_self_ref_order`, which is the subtle half of the ordering contract.
- **`StoreModelSourceStreamTestCase` runs against real `Store` rows and the real registry.** `test_stream__skip_errored` covering both `NULL` and `""` is the historical case `NullIf` exists for, and `test_stream__models_in_dependency_order` pins the registry-order guarantee at the level a consumer actually depends on.
- **`StoreManager(models.Manager.from_queryset(StoreQueryset))`** — deletes a hand-rolled `get_queryset` and makes every queryset method reachable from the manager, which is what let `exclude_has_deserialization_error()` read as one call at the use site.
- **The `NullIf`-over-`OR` comment** (`core.py:449-450`) records *why* the odd-looking annotation is there, on both backends. That is the comment that stops someone "simplifying" it back into an `OR`.
- **Gating the `_self_ref_order` sort on `get_self_referential_fk`** (`registry.py:121`) keeps the majority of models off an unindexed sort while preserving the tree ordering where it is needed.

---

## 3. Reviewer Guidance, Answered

> Do the tests cover the new and refactored code enough?

Coverage of the *new* code is good, and the real-row tests are the right choice. Two gaps, both listed above: `__slots__` effectiveness, and unfiltered `AppModelSource` dedup (whose old test moved to the filtered path).

> Did I properly translate the self-ref-ordering and deserialization error filtering?

Yes on both, as far as this PR's surface goes.

- **Error filtering** is a faithful extraction of `operations.py:326-336`, and the `TextField()` output field is more accurate than the `BooleanField()` it came from. Both null and `""` are handled, and tested against real rows.
- **Self-ref ordering** matches the semantics `_update_legacy_self_ref_order_for_model` establishes (`operations.py:728-745`): roots at `0`, children at parent+1, `NULL` when unresolvable — so ascending-nulls-last is parents-before-children with the unresolvable tail last. Correct.

One difference to carry forward rather than fix here: `get_store_querysets` scopes each queryset to a single `model_name`, where `_deserialize_from_store` ORs in `morango_model_dependencies` (`operations.py:307-311`). That is the better shape — the registry already orders dependencies first via `_insert_model_in_dependency_order`, so each record is now covered exactly once instead of revisited — and it matches `get_model_querysets`. Worth knowing when the consumer lands, since the legacy self-ref branch computed `clean_parents`/`dirty_children` over that wider set.

The other thing the consumer will have to supply: the legacy self-ref branch didn't just order records, it *classified* the failures — `MorangoDirtyParent` and `MorangoMissingParent` (`operations.py:380-396`). Ordering by `_self_ref_order` replaces the iteration but not the diagnostics.

---

## 4. Next Steps

- [ ] Make `stream()` safe or self-diagnosing without `begin()` (`source.py:47`)
- [ ] Decide on stateless prefix-exclusion vs. keeping `_seen`, and record which and why
- [ ] Document or narrow `begin()`'s clearing of an injected `fk_cache`
- [ ] Optional: `_filter_deserialization_error` rename; `__slots__` regression test; cross-partition-FK invariant in the comment; drop redundant `abc.ABC`
- [ ] Confirm docs/CHANGELOG are intentionally deferred to a later part of the series

tests/testapp/tests/sync/stream/test_source.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_seen_is_uninitialized_before_begin(self):
+        source = FakeSource("test")
+        self.assertIsNone(source._seen)
+
+    def test_begin_initializes_seen(self):
+        source = FakeSource("test")
+        source.begin()
+        self.assertEqual(source._seen, set())
+
+    def test_begin_resets_seen_between_runs(self):
```
rtibblesbot, line 39: **praise:** Pins the reuse contract rather than just the fix — streams the same source twice and says in the comment why run two would otherwise yield nothing. Fails loudly if `_seen` ever moves back into `__init__`.

rtibblesbot (commented):
PR #318 — 12 of 15 prior findings resolved; 3 answered by @bjester and left as-is. No new findings. CI passing; ran `tests/testapp/tests/sync/stream/`, `test_registry.py`, `models/test_core.py` locally (119 passed).

Still open, all deferred to the author's answers — recorded, not blocking:

- `morango/sync/stream/source.py:84` — partition-major iteration drops model-dependency ordering. Answered: no cross-partition FKs, so ordering within a partition suffices.
- `morango/sync/stream/source.py:67` — prefixes already covered by a shorter prefix still get a full pass. Answered: supplying a reduced filter is the caller's job.
- `morango/models/core.py:444` — `operations.py:326-336` still hand-rolls the annotation the new helper extracts. Answered: that code is slated for removal.

RESOLVED — morango/sync/stream/source.py:11 — `SourceTask` declares no `__slots__`
RESOLVED — morango/sync/stream/source.py:51 — `_seen` accumulates when nothing can duplicate
RESOLVED — morango/sync/stream/source.py:51 — `_seen`/`fk_cache` not reset in `begin()`
RESOLVED — morango/sync/stream/source.py:73 — "passes thoughts to `stream_for_filter`"
RESOLVED — morango/sync/stream/deserialize.py:23 — fresh `{}` per store record gives the FK cache no reuse
RESOLVED — morango/sync/stream/deserialize.py:57 — `skip_errored` default inverted the legacy default
RESOLVED — morango/sync/stream/deserialize.py:69 — `fk_cache or {}` drops a caller-supplied empty dict
RESOLVED — morango/registry.py:122 — `_self_ref_order` sort applied to every model
RESOLVED — morango/models/core.py:453 — `NullIf` declared with `BooleanField` output
RESOLVED — tests/testapp/tests/sync/stream/test_deserialize.py:57 — `StoreModelSourceTestCase` mocks the ORM throughout
RESOLVED — tests/testapp/tests/sync/stream/test_deserialize.py:138 — praise: NULL and empty-string `deserialization_error` against real rows
ACKNOWLEDGED — tests/testapp/tests/test_registry.py:130 — praise: asserts returned rows rather than generated SQL
CONTESTED — morango/sync/stream/source.py:84 — partition-major iteration drops model-dependency ordering
CONTESTED — morango/sync/stream/source.py:67 — prefixes covered by a shorter prefix still get a pass
CONTESTED — morango/models/core.py:444 — `operations.py` call site still hand-rolls the annotation
</details>

---

Ran an automatic code-only delta review triggered by new commits on a previously reviewed PR:
- Core review pass only — specialized frontend/backend lenses and manual QA run when a review is explicitly requested

</details>

morango/models/core.py
```diff
@@ -440,10 +441,30 @@ def char_ids_list(self):
             .values_list("fixed_id", flat=True)
         )
 
+    def filter_deserialization_error(self, has_error: bool) -> "StoreQueryset":
+        """
+        Filters the queryset to return Store records that have or have not any
+        deserialization errors
+        """
+        # this nullIf assertion is generally more performant than an OR'd statement on
```
rtibbles, line 449: So, this is because the deserialization_error can be _either_ empty string or null to indicate no error?

morango/sync/stream/deserialize.py
```diff
@@ -0,0 +1,92 @@
...
+        self,
+        profile: str,
+        sync_filter: Optional[Filter] = None,
+        dirty_only: bool = True,
+        partition_order: str = "asc",
+        fk_cache: Optional[Dict] = None,
+        skip_errored: bool = False,
+    ):
+        """
+        :param profile: The Morango model profile
```
rtibbles, line 61: I wonder what the way to use type hints but maintain full docstring readability is? Without the type hints, the parameter types would be recorded in the docstring, but with them, that is, rightly, redundant.

Ah, apparently the way to move the definitions inline is to use `Annotated` https://docs.python.org/3/library/typing.html#typing.Annotated (which was added in 3.9, so we can't use yet anyway!)

morango/sync/stream/deserialize.py
```diff
@@ -0,0 +1,92 @@
...
+        for store_qs in syncable_models.get_store_querysets(self.profile):
+            qs = store_qs
+            if partition_condition is not None:
+                qs = qs.filter(partition__startswith=partition_condition)
+            if self.dirty_only:
+                qs = qs.filter(dirty_bit=True)
+            if self.skip_errored:
+                qs = qs.exclude_has_deserialization_error()
+
+            for store_model in qs.iterator():
```
rtibbles, line 91: This is not a suggestion - just wondered if you had looked at how much of the memory overhead comes from Django model inflation - less of a concern if we're only going model by model, but just wondered if you'd measured its contribution, as it's one of the main things that reduces memory consumption for ValuesViewsets by not inflating the models.

morango/sync/stream/source.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+T = TypeVar("T", bound=SourceTask)
+
+
+class MorangoSource(Source[T], abc.ABC):
+    """
+    Common source functionality for Morango sources, such as SyncableModels and Store records.
+    """
+
```
rtibbles, line 29: Any need for `__slots__` here, so that all classes inherit them by default, but can be overridden when needed?

tests/testapp/tests/sync/stream/test_source.py
```diff
@@ -0,0 +1,50 @@
...
+        """A source may be streamed more than once, and must not carry state across runs"""
+        source = FakeSource("test", sync_filter=Filter("a"), task_ids=("1", "2"))
+
+        source.begin()
```
rtibbles, line 43: Just wondering if having to explicitly `begin` the source is helpful, or whether it should just automatically `begin` when you `stream`? Is there a case where stream is called twice in a row without a begin in between?

rtibbles (commented):
Nothing jumped out at me as a concern - changes make sense, and tests show good coverage of scenarios - just a few questions/observations.

morango/models/core.py
bjester: Yes, an empty string was used previously to indicate no error.

morango/sync/stream/deserialize.py
bjester: Well I view these as different than `:type profile: str`. These are comments regarding the arguments, which are not required to explicitly mention the type but might refer to it to describe the usage.

morango/sync/stream/deserialize.py
bjester: I think it's worth investigating, but a decent amount of Morango's serialization / deserialization machinery relies on model objects, so it would require a completely separate refactor. It might be more beneficial for time performance than memory.

morango/sync/stream/source.py
bjester: It doesn't hurt. I'll add it.

tests/testapp/tests/sync/stream/test_source.py
bjester: I don't think the use case of `begin` is the responsibility of `stream`, and so keeping them separate keeps that clear. The `Pipeline` implements the overall functionality (essentially the contract of this API), and although it calls one immediately after the other-- meaning they could be consolidated-- the structure would not lend the base source's behavior to be overridden later if need be.

There is no current use case for reusing streams, but I did envision potentially reusing a source if the entire sync is pipelined (from client to server).

rtibbles (approved):
Good to merge with or without slots as to your preference.

PR #322, opened 2026-04-29
Implement self ref fk ordering

### LINKED ISSUE #297: Tracking Self-Referential FK Order
**Target branch:** release-0.9.x

## Current behavior

The current deserialization behavior recurses into 'children' models when it encounters a model with a self-referential foreign key (FK). This contributes to its monolithic nature and its memory consumption by requiring the algorithm to track more data in memory to ensure comprehensive deserialization of store data.

The serialization process tracks the FK value for self-referential models in a field called `_self_ref_fk`. This allows it to lookup those FK records using that value.

Additionally, the Morango registry, which tracks the syncable models defined by any apps using Morango, orders them according to their cross-model FK dependencies.

## Desired behavior

The deserialization algorithm would be more amenable to the streaming architecture by ordering the store records more intelligently before they enter the pipeline. Combined with the existing ordering by the Morango registry of syncable models, this should create a predictable order in which store records can be properly deserialized without broken FKs.

#### Data Model
This specifically addresses self-referential FK ordering within the same model (e.g., parent-child relationships in hierarchical data). The registry handles cross-model dependencies, but self-referential FKs need special ordering.
- A new nullable field `_self_ref_order` should be added to the `AbstractStore` with type of `Integer`
- The new field be greater than or equal to `0` unless null
- Django model migrations should be created to add the new field to `Store` and `Buffer`
- The `SyncableModel` should have a new tuple attribute `morango_ordering` which should hold one or more field expressions in the format that Django uses for `order_by()` (similar to `Meta.ordering`)
- The `SyncableModelRegistry.get_model_querysets` should order its querysets according to `morango_ordering` if set; **nulls should always be last**

#### Serialization
Update `StoreUpdate` transform in [serialize.py](https://github.com/learningequality/morango/blob/release-v0.9.x/morango/sync/stream/serialize.py) to set `_self_ref_order` during `Store` record creation and updates.
- The `_handle_store_create` method should be updated:
  - if the model is not self-referential, the order field should be null
  - if `_self_ref_fk` is `None` and the model is self-referential, the order field should be `0`
  - otherwise, it should query for the parent's order (the `Store._self_ref_order` for `id=_self_ref_fk`) 
- The `_handle_store_create` method should be updated:
  - to detect if the `_self_ref_fk` has changed, and if so, update the order like `_handle_store_create`

#### Compatibility
Morango should communicate during a sync whether it supports ordering with `_self_ref_order`. Later on, this will allow Morango to determine whether it can fully streamline the sync.
- A new capability `SELF_REF_ORDER` should exist in `constants/capabilities.py`
- The capability should be provided by default during a sync
- If the client device sends the capability, the server should trust and expect `_self_ref_order` in the sent data
- If the client does not send the capability, the server should compute the order during the `dequeue` stage

## Value add

Provides more certainty to the pipeline that we process store records in the order of their relational dependency in the DB.

## Possible tradeoffs

- Morango defines other fields on `SyncableModel` for consumers to override, which could be better to have in the model `Meta`. For example, you can define `ordering` in the model `Meta`, but we're putting `morango_ordering` on the model itself.
- Morango could let the app models provide the ordering information through a defined API, but backwards compatibility would still need logic to determine it manually. Therefore always leveraging that logic seems sensible, unless we let go of backwards compatibility
- We're adding yet another capability. While it isn't a concern at the moment, growth of the capabilities should be monitored since it's sent in an HTTP header, which has inconsistent limits on size depending on server.

## AI usage
AI was used for brainstorming approaches.

ozer550:
## Summary
Introduce deterministic ordering support for self-referential foreign key (FK) models during sync by adding a _self_ref_order field to Morango store records, allowing store/buffer rows to be processed parent-before-child in a streaming-friendly way. Adds a morango_ordering hook for syncable models so Morango can order querysets consistently (with nulls last), and adds a SELF_REF_ORDER capability to negotiate support between peers and preserve backwards compatibility.

## TODO

## Reviewer guidance

## Issues addressed

closes #297 

## Documentation


morango/sync/stream/serialize.py
```diff
@@ -200,11 +207,29 @@ def _handle_store_create(self, task: SerializeTask):
...
-            self_ref_fk_value = getattr(task.obj, self_ref_fk)
-            kwargs["_self_ref_fk"] = self_ref_fk_value or ""
+            self_ref_fk_value = getattr(task.obj, self_ref_fk) or ""
+            kwargs["_self_ref_fk"] = self_ref_fk_value
+            kwargs["_self_ref_order"] = self._compute_self_ref_order(self_ref_fk_value)
 
         task.set_store(Store(**kwargs))
 
+    @staticmethod
+    def _compute_self_ref_order(self_ref_fk_value):
```
bjester: So from our discussion on slack and seeing this, I think some of the behaviors have been taken very literally, which is my fault for not making that clearer.

> if _self_ref_fk is None and the model is self-referential, the order field should be 0
otherwise, it should query for the parent's order (the Store._self_ref_order for id=_self_ref_fk)

We might need to query the database for this, but I think it's worth thinking through how and when this should occur. This satisfies the criteria of the behavior, but the engineering choices are up to you. The behaviors are not meant to be taken literally that this functionality should explicitly be placed here.

I say this for a couple reasons:
1. Look at how we manage database calls-- in `StoreLookup` we do bulk queries for the store records, and how `StoreUpdate` actually does none. There's room for improvement in managing these queries.
2. The circumstances may not always work with the pipeline because it buffers records.  On this note, I'm thinking in regards to this test case `test_handle_store_create__self_ref_parent_not_in_store` and whether that is an acceptable outcome.

More on (2): say we have our buffer batch size set to 2. In the first batch, we have a parent and a child. Since the batch size is two, the parent won't be saved yet when the child is processed, so this function should result in `None`. Then lets say batch #2 has two more children, of the first child (grandchildren). The function should find the store, but the store would have a `None` order, leading to `None` for the grandchildren. Thus we have a parent-child relationship which could get out of order because of `None`.

morango/sync/stream/serialize.py
```diff
@@ -138,6 +162,84 @@ def transform(self, tasks: List[SerializeTask]) -> List[SerializeTask]:
...
+                    continue
+                child_order = parent_order + 1
+                task.set_self_ref_order(child_order)
+                child_id = task.obj.id
+                known_order_by_id[child_id] = child_order
+                queue.append(child_id)
+
+        return tasks
+
+
```
bjester, line 257: This looks pretty good, but I think it can be simplified a little more.

First, we know there's a chance of at least two loops:
1. Your first loop, that goes over tasks and sets self-ref properties, and determines what needs further lookups
2. Another loop to process the parent lookups

With the `morango_ordering` feature, we expect that a parent comes before a child in the stream, if the parent is being serialized. That means, in the first loop, we can eagerly check the cache for the parent. If the parent is not in the cache, we queue the child for further lookups. That means we're able to handle all children, whose parents are in the stream, in the first loop, minimizing the length of the second loop.

The lookup of parents in the store looks good, and setting those parents in the cache makes sense.

As written, the second loop goes over all cached parent IDs. This could be significantly larger than the remaining children queued by the first loop. In other words, looping over the remaining queued children directly should be simpler and because the cache is indexed by parent, we can easily lookup the parent order using the `self_ref_fk_value` already set on the child in the first loop. Altogether, this should eliminate the need for `children_by_parent`, and simplifies the `external_parent_ids` to the parent IDs pulled from the remaining queued children.

As it's written, the second loop will get slower and slower as the stream count increases. This is especially so since the cache isn't reset when we detect those 'checkpoints' in the stream where we go from one model to another. But simplifying the second loop as I mentioned will eliminate the former cache issue, leaving only the latter (to reset the cache at the appropriate time)

morango/sync/operations.py
```diff
@@ -721,7 +725,41 @@ def _queue_into_buffer_v2(transfersession, chunk_size=200):
...
+def _update_legacy_self_ref_order_for_model(queryset):
+    # root nodes set the _self_ref_order to 0
+    queryset.filter(_self_ref_fk="").exclude(_self_ref_order=0).update(_self_ref_order=0)
+    # reset the _self_ref_order to None for all records that have a parent
+    queryset.exclude(_self_ref_fk="").exclude(_self_ref_order=None).update(
+        _self_ref_order=None
+    )
+
+    parent = Store.objects.filter(
+        id=Cast(OuterRef("_self_ref_fk"), UUIDField()),
```
bjester, line 737: I added the `Cast` here, as postgres complained about a mismatch in type.

bjester (approved):
I approve of @ozer550's changes! 

As for my changes, I resolved merge conflicts, fixed a postgres error, ensured our buffered chunks are always partitioned by model, and did some refactoring for partitioned buffers and self-ref FK lookups.

rtibbles (approved):
No concerns from a read through, and the implementation matches the spec to my eyes. No obvious artifacts from the merge conflict resolution.

PR #331, opened 2026-06-04
Remove multiprocessing fallback for key generation. Stay in process.

rtibbles:
## Summary

Removes the multiprocessing usage when creating keys using pure Python RSA - when no cryptography library is installed.

## Reviewer guidance

I don't think this loses any error handling as we were calling this code in the `except` clause before anyway.

## Issues addressed

No explicit issues, may clean up zombies on no cextension Kolibris though.


bjester (approved):
LGTM

PR #336, opened 2026-06-12
Cleanup from UV migration + standardization of workflows and package config

bjester:
## Summary
- Restructures dependencies in `pyproject.toml`, making `dev` and `test` dependency groups instead of optional, leaving `postgres` and `accelerated` as optional
- Updates instructions and workflows to install groups as necessary
- For python < 3.9 where we use tox < 4, it overcomes lack of dependency group support by using `uv` to install the group deps into the system packages, then allows tox to use those
- Reorganizes skip checks to the job level instead of step level
- Adds concurrency control to cancel in-progress actions when PR is updated
- Adds action linting
- Fixes issues flagged by action linting, except one that is ignored
- Updates dependabot cadence to check monthly instead of weekly
- Updates import lint/formatting rule to enforce single-line imports
- Reformats python files due to above change
- Adds fan-in summary job for a single required PR check
- Ensures all actions use full explicit versions, except duplicate check (TODO: https://github.com/learningequality/.github/issues/83)

## Reviewer guidance
- Code reformatting looks okay?
- CI passes?

## AI Usage
Had AI start the process, then wrapped it up myself, as there were challenges with python+tox+pip versioning and other stuff was just copy & paste.


.github/workflows/tox.yml
```diff
@@ -227,23 +232,26 @@ jobs:
...
-        if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
         run: |
           python -m pip install --upgrade pip
           pip install tox
       - name: Test with tox
-        if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
         run: |
           $wheel = Get-ChildItem dist\*.whl | Select-Object -ExpandProperty FullName -First 1
-          tox --installpkg $wheel -e windows
+          tox --installpkg $wheel -e windows --sitepackages
```
rtibbles, line 261: Would it be helpful to add a fan-in job here like I did on the kolibri tests?

rtibblesbot (commented):
Clean tooling modernisation — migrating `test`/`dev`/`docs` to PEP 735 `[dependency-groups]`, upgrading pre-commit hooks, and improving CI hygiene. All 28 CI checks pass.

No UI files changed; Phase 3 skipped.

Two nitpicks: a pair of redundant step-level skip conditions that the refactor missed cleaning up in `check_migrations_sqlite.yml`, and inconsistent action-version pin style in the new `pre-commit.yml`. Everything else looks solid.

**backend review:** This PR is a clean, mechanical refactor: the vast majority of changes are import statement reformatting enforced by ruff's new `force-single-line = true` isort setting, applied consistently to all ~80 Python files. No Python backend logic was changed — every diff in `morango/` and `tests/` is either a grouped-import split or a trivial formatting tweak (e.g. `_self_ref_order` field definition line-length). The `.git-blame-ignore-revs` entry correctly marks the reformatting commit.

The more substantive changes are in the build and CI layer: `test`/`dev`/`docs` groups migrate from `[project.optional-dependencies]` to PEP 735 `[dependency-groups]`, which is the right separation (they're internal tooling deps, not published package extras). `postgres` and `accelerated` correctly remain as `[project.optional-dependencies]`. The `dev` group gains `{include-group = "test"}`, making it a proper superset. CI workflows gain concurrency-cancel blocks and move per-step `if: skip_check` guards to the job level, which is cleaner. EOL Python matrix entries gain `--sitepackages` + `uv pip install --system --group test` to work around uv's limitations with Python < 3.8.

Two minor issues found: an unpinned third-party action in the new pre-commit workflow, and two leftover step-level `if` conditions in `check_migrations_sqlite.yml` that are now redundant.

**[nitpick]** Redundant step-level skip conditions in migration check workflow

The `migration_test` job in `check_migrations_sqlite.yml` now has a job-level `if: ${{ needs.pre_job.outputs.should_skip != 'true' }}` (added by this PR), which means the job is skipped entirely when `should_skip` is true. Two step-level conditions that pre-date this PR were not cleaned up:

- "Install build dependencies" step (line 55) still has `if: ${{ needs.pre_job.outputs.should_skip != 'true' }}`
- "Run migrations" step (line 99) still has the same condition

These are now no-ops — the job never reaches them when skipping. All other workflows had their step-level conditions promoted to the job level correctly; these two were missed. Safe to remove both.

**[praise]** EOL Python workaround is well-crafted

The approach for Python < 3.9 (tox < 4, which doesn't support `dependency_groups`) — pre-installing test deps into the system Python via `uv pip install --system --group test` and then passing `--sitepackages` to tox — is a clean, minimal workaround with no footguns. It avoids duplicating the dependency list and stays consistent with how modern Python versions get their deps. The same pattern is applied uniformly across EOL unit tests, EOL cryptography tests, Windows, and migration checks, and CI confirms it works across Python 3.6–3.8.

**[praise]** Skip-condition refactor improves CI clarity

Moving `if: ${{ needs.pre_job.outputs.should_skip != 'true' }}` from every individual step up to the job level is the correct pattern: it prevents GitHub from even starting the job (saving runner minutes and reducing log noise) rather than running the job and skipping each step individually. The change is applied consistently across `tox.yml`, `check_migrations_sqlite.yml`, and `check-docs.yml`.

**[praise]** git-blame-ignore-revs housekeeping

Adding the import-reformatting commit to `.git-blame-ignore-revs` is good housekeeping — `git blame --ignore-revs-file .git-blame-ignore-revs` (and GitHub's blame UI, which respects this file) will skip over the mechanical reformatting and show the original authors.

**[praise]** PEP 735 dependency groups migration is correct

`test`, `dev`, and `docs` moving from `[project.optional-dependencies]` to `[dependency-groups]` is the right call: these are internal tooling deps that shouldn't be exposed as published package extras, and uv's native group support is the proper mechanism. `postgres` and `accelerated` correctly remain as optional-dependencies since they change runtime behaviour for consumers. The `{include-group = "test"}` entry in `dev` is a clean DRY improvement.

**[praise]** Import reformatting handled cleanly end-to-end

The `force-single-line = true` isort rule is introduced in `pyproject.toml`, applied mechanically by ruff, and the reformatting commit is registered in `.git-blame-ignore-revs`. Each import is now independently navigable in IDEs and diffs are more granular. The scope is complete — all Python files in the diff follow the new style.

Comments on lines not in diff:

**.github/workflows/check_migrations_sqlite.yml:55** — **[nitpick]** Redundant step-level if conditions in check_migrations_sqlite.yml

The `migration_test` job gained a job-level `if: ${{ needs.pre_job.outputs.should_skip != 'true' }}` in this PR (line 49), which correctly prevents the entire job from running when the skip check fires. However, two step-level copies of the same condition were not removed:

- Line 55: `Install build dependencies` step
- Line 99: `Run migrations` step

All other equivalent step-level conditions were removed in this PR. These two are now dead code — the job won't reach them when `should_skip == 'true'`. Low impact, but inconsistent with the rest of the cleanup.


---

- Discarded findings citing files outside the diff; checked CI status and linked issue acceptance criteria
- Calibrated the final verdict deterministically from finding severities, CI status, and QA evidence

</details>

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,43 @@
...
+          github_token: ${{ github.token }}
+          paths_ignore: '["**.po", "**.json"]'
+
+  linting:
+    name: All file linting
+    needs: pre_job
+    if: ${{ needs.pre_job.outputs.should_skip != 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6.0.2
```
rtibblesbot, line 35: **[nitpick]** Action version pin style inconsistency in pre-commit.yml

The new `pre-commit.yml` pins actions to full semver (`actions/checkout@v6.0.2`, `astral-sh/setup-uv@v8.2.0`), while every other workflow in this repo uses major-version-only pins (`@v6`, `@v7`). The two approaches have different trade-offs (full pins are more reproducible; major-only picks up patch/minor fixes automatically) but mixing them in the same repo adds maintenance friction — e.g., a bump PR that updates `@v7` → `@v8` elsewhere would also need to know to update `@v8.2.0` here.

Also worth noting: `setup-uv@v8.2.0` is a different major than `@v7` used elsewhere; if intentional, a brief comment would help future maintainers understand why.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,43 @@
...
+    name: Path match check
+    runs-on: ubuntu-latest
+    # Map a step output to a job output
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          github_token: ${{ github.token }}
```
rtibblesbot, line 26: **[suggestion]** Unpinned third-party action in new pre-commit workflow

`.github/workflows/pre-commit.yml` (added by this PR) uses `fkirc/skip-duplicate-actions@master` — a floating branch reference. If the upstream repo pushes a malicious or breaking commit to `master`, all workflows using this action are affected immediately. GitHub's security hardening guide recommends pinning third-party actions to a full-length commit SHA (e.g. `fkirc/skip-duplicate-actions@f7549b5ca11b7f3af12f08a7e3a2d5b5c3e1e3a`). The same pattern exists in the pre-existing `tox.yml` and `check-docs.yml`, so this is a pre-existing issue being replicated to the new file; worth addressing across all workflows.

rtibbles (commented):
Just one question - I didn't read the reformatting commit to save my brain, just assuming that ruff works as expected :)

.github/workflows/tox.yml
bjester: Yeah that's a helpful feature here. I can add that too

.github/workflows/pre-commit.yml
bjester: We'll be tackling this wholesale across all our LE repos, so I'll leave it to that: https://github.com/learningequality/.github/issues/83

rtibblesbot: Understood — deferring to the wholesale LE-wide initiative. Thanks for the link.

.github/workflows/pre-commit.yml
bjester: Added explicit versions to all except the duplicate checking action.

.github/workflows/tox.yml
bjester: Updated

rtibbles (approved):

.github/workflows/pre-commit.yml
rtibblesbot: Thanks for adding the explicit versions — that addresses the consistency concern.

PR #341, opened 2026-06-26
Add deferrable FK utility, test case, and migration

bjester:
## Summary
- Adds utility for updating existing FK constraints, that are immediate by default, to have a deferrable flag
  - This corrects an issue where databases that were created prior to Django 3.1 did not set the deferrable flag, and after that, Django enabled FK checks by default. Django relies on deferrable FK constraints for its cascade deletion because it executes deletions out of order (as shown in the issue).
  - The utility works for both a Django migration and manual invocation, such that Kolibri can include it as a version upgrade migration.
- A Django migration was added to invoke the utility explicitly for morango models automatically
- A test verifies the proof-of-concept of how the utility operates, including its preservation of data

Creation of this utility was started in Kolibri, when it was realized other apps (like Morango) may need this. Since Morango is a standalone package with respect to Kolibri, the utility was added here, made to automatically handle morango models, and to allow Kolibri to leverage the utility in its own way.

## Reviewer guidance
- Are the tests sufficient?
- Is there test coverage lacking?

I ran Kolibri 0.16.2 to generate an old database with the issue, along with some data (classes and users). With these changes integrated into Kolibri, I then ran both via a Kolibri upgrade on top of the the 0.16.2 home, to verify that it updated the FK constraints. All data was preserved.

I can provide the 0.16.2 home upon request.

### Before (0.16.2)
```
sqlite> .schema kolibriauth_collection
CREATE TABLE IF NOT EXISTS "kolibriauth_collection" (
  "id" char(32) NOT NULL PRIMARY KEY, 
  "_morango_dirty_bit" bool NOT NULL,
   "_morango_source_id" varchar(96) NOT NULL, 
  "_morango_partition" varchar(128) NOT NULL, 
  "name" varchar(100) NOT NULL, 
  "kind" varchar(20) NOT NULL, 
  "dataset_id" char(32) NOT NULL REFERENCES "kolibriauth_facilitydataset" ("id"), 
  "parent_id" char(32) NULL REFERENCES "kolibriauth_collection" ("id")
);
```

### After (pending 0.19.5)
```
sqlite> .schema kolibriauth_collection
CREATE TABLE IF NOT EXISTS "kolibriauth_collection" (
  "id" char(32) NOT NULL PRIMARY KEY, 
  "_morango_dirty_bit" bool NOT NULL, 
  "_morango_source_id" varchar(96) NOT NULL, 
  "_morango_partition" varchar(128) NOT NULL, 
  "dataset_id" char(32) NOT NULL REFERENCES "kolibriauth_facilitydataset" ("id") DEFERRABLE INITIALLY DEFERRED,
  "name" varchar(100) NOT NULL, 
  "kind" varchar(20) NOT NULL, 
  "parent_id" char(32) NULL REFERENCES "kolibriauth_collection" ("id") DEFERRABLE INITIALLY DEFERRED
);
```

## Issues addressed

Supports resolution of https://github.com/learningequality/kolibri/issues/14884

## AI Usage
Creation of this utility was started in Kolibri with AI. It generated the bulk of the utility and the test, then it was brought over to Morango, and a human refactored the utility to make it flexible for both Morango and Kolibri, and to isolate looping over applications.


morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,127 @@
...
+        self,
+        include_app_labels: Optional[List[str]] = None,
+        exclude_app_labels: Optional[List[str]] = None
+    ):
+        self.include_app_labels = include_app_labels
+        self.exclude_app_labels = exclude_app_labels
+
+    def _iter_apps(self, apps: Apps):
+        for app_config in apps.get_app_configs():
+            if self.exclude_app_labels is not None and app_config.name in self.exclude_app_labels:
```
rtibblesbot: **suggestion:** The parameters are named `include_app_labels`/`exclude_app_labels`, which in Django parlance refers to the short `AppConfig.label` (e.g. `auth`, `contenttypes`). The implementation compares against `app_config.name` — the full Python module path (e.g. `django.contrib.auth`). For `morango` and `facility_profile` the two values happen to be the same string, so all current callers work. But a Kolibri caller who passes the label (e.g. `kolibriauth` for the app whose name is `kolibri.core.auth`) would silently match nothing — no error, no log, migration is a no-op.

Options: rename the parameters to `include_app_names`/`exclude_app_names` to match what is compared, or switch the comparison to `app_config.label` (and update `BASE_APPS` in the test, which currently uses full dotted paths).

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,127 @@
...
+                if not _table_has_immediate_foreign_key(conn, table_name):
+                    continue
+                logger.info(
+                    "Rebuilding table %s to make foreign key constraints deferrable",
+                    table_name,
+                )
+                # Rebuilding the table from the current model state regenerates the
+                # column definitions, which SQLite always emits with the deferrable
+                # clause (see DatabaseSchemaEditor.sql_create_inline_fk).
+                schema_editor._remake_table(model)
```
rtibblesbot: **suggestion:** `_remake_table` is a private Django method (leading underscore, `sqlite3.schema.DatabaseSchemaEditor`). It has changed signature in the past, and there is no public alternative that rebuilds DDL from model state. The existing comment (lines 116–118) is good; it would be worth adding a note on which Django versions this has been tested against, and that `DatabaseSchemaEditor.sql_create_inline_fk` is the reference point to check if a future Django upgrade breaks it.

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    Determine whether ``table_name`` has at least one foreign key constraint that
+    is not deferrable. Tables are created with all constraints in the same style,
+    so the presence of a ``REFERENCES`` clause without ``DEFERRABLE`` indicates an
+    immediate constraint that needs to be rebuilt.
+    """
+    sql = _get_table_sql(conn, table_name)
+    if not sql:
+        return False
+    return "REFERENCES" in sql and "DEFERRABLE" not in sql
```
rtibblesbot: **suggestion:** This check is table-level: if `DEFERRABLE` appears *anywhere* in the DDL — even for just one FK added under Django 3.1+ — the function returns `False` and the repair pass skips the table. A table with mixed immediate and deferrable FKs (e.g. created pre-3.1 then extended via a later migration) would be silently skipped, leaving its original immediate constraints unrepaired. This edge case is unlikely in practice; worth noting as a known limitation in the docstring.

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,127 @@
...
+                # column definitions, which SQLite always emits with the deferrable
+                # clause (see DatabaseSchemaEditor.sql_create_inline_fk).
+                schema_editor._remake_table(model)
+
+    def run(self):
+        """
+        Invoke this utility outside of Django migrations, manually passing in Django's app registry
+        and schema editor.
+        """
+        with connection.schema_editor() as schema_editor:
```
rtibblesbot: **suggestion:** `connection.schema_editor()` defaults to `atomic=True`, while the integration test uses `connection.schema_editor(atomic=False)`. `_remake_table` is not known to be incompatible with a wrapping transaction, so the default is probably fine — a brief comment here would prevent a future reader from wondering whether the discrepancy was intentional.

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,127 @@
...
+        which will make FKs deferrable
+        """
+        conn = schema_editor.connection
+        # these should be in sync, but since we call `_remake_table`, specific to this schema editor
+        # class, we ensure we have the correct instance to begin with
+        if conn.vendor != "sqlite" or not isinstance(schema_editor, DatabaseSchemaEditor):
+            return
+
+        for app_config in self._iter_apps(apps):
+            for model in app_config.get_models(include_auto_created=True):
```
rtibblesbot, line 110: **praise:** `include_auto_created=True` ensures M2M through-tables — which carry FK columns on both sides — are included in the repair pass. Without this, an entire class of tables would be silently skipped.

tests/testapp/tests/test_deferrable_foreign_keys.py
```diff
@@ -0,0 +1,87 @@
...
+        with connection.schema_editor(atomic=False) as schema_editor:
+            op(global_apps, schema_editor)
+
+        # now deferrable again, and data preserved
+        self.assertIn("DEFERRABLE", _get_table_sql(connection, table))
+        self.assertTrue(Facility.objects.filter(id=f_id).exists())
+
+    def test_iter_apps__all(self):
+        op = MakeForeignKeysDeferrable()
+        self.assertEqual([app.name for app in op._iter_apps(global_apps)], [
```
rtibblesbot, line 70: **nitpick:** Asserting the complete ordered list of all installed apps couples this test to `INSTALLED_APPS` ordering rather than the filtering logic under test. Adding or reordering an app in test settings would break it silently. Checking that specific expected apps are present (or absent) would be less brittle.

tests/testapp/tests/test_deferrable_foreign_keys.py
```diff
@@ -0,0 +1,87 @@
...
+        immediate = sql.replace(" DEFERRABLE INITIALLY DEFERRED", "")
+        assert "DEFERRABLE" not in immediate
+        with connection.cursor() as cursor:
+            cursor.execute(
+                "SELECT sql FROM sqlite_master WHERE type='index' AND tbl_name=%s AND sql IS NOT NULL",
+                [table],
+            )
+            indexes = [r[0] for r in cursor.fetchall()]
+            cursor.execute("PRAGMA foreign_keys = OFF")
+            cursor.execute("DROP TABLE {}".format(table))
```
rtibblesbot: **nitpick:** String-formatting a table name into DDL is a habit worth avoiding even when the value is safe. `connection.ops.quote_name(table)` handles quoting correctly and makes the intent explicit. (Parameterised placeholders cannot be used for identifiers, so this is the right alternative.)

morango/migrations/0004_deferrable_foreign_keys.py
```diff
@@ -0,0 +1,24 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("morango", "0003_store_deserialization_errors"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            MakeForeignKeysDeferrable(include_app_labels=["morango"]),
+            migrations.RunPython.noop,
```
rtibblesbot, line 22: **praise:** `RunPython.noop` is the correct reverse here — there is no meaningful way to un-deferrable a constraint without the original schema. More honest than a placeholder that claims to undo something it can't.

rtibblesbot (commented):
Solid implementation — checking `sqlite_master`, calling `_remake_table`, and no-oping on non-SQLite backends is clean. CI's only failure is the contributor gate, not test failures.

Findings are suggestions and nitpicks. The parameter-naming issue has a real impact on Kolibri callers; the rest are commentary or test hygiene.

- **suggestion:** `*_app_labels` params compare against `app_config.name`, not Django's label — silent mismatch for Kolibri callers
- **suggestion:** `_remake_table` private-API dependency should be documented
- **suggestion:** mixed immediate/deferrable FK edge case — note as known limitation
- **suggestion:** `atomic=True` in `run()` vs `atomic=False` in test — worth a clarifying comment
- **nitpick:** `test_iter_apps__all` asserts full ordered list; `RunPython.noop` reverse is the right call

---

</details>

tests/testapp/tests/test_deferrable_foreign_keys.py
bjester: I'm not worried about this. The test suite uses a test app, so the `INSTALLED_APPS` are unlikely to change often. Any test failure when changed is easily correctable.

rtibblesbot: Fair point — the test environment is controlled and any breakage would be obvious and easy to fix. Withdrawing this finding.

morango/deferrable_foreign_keys.py
bjester: Fixed

morango/deferrable_foreign_keys.py
bjester: Added a comment

morango/deferrable_foreign_keys.py
bjester: Fixed

morango/deferrable_foreign_keys.py
bjester: Added a comment

tests/testapp/tests/test_deferrable_foreign_keys.py
bjester: Fixed

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,136 @@
...
+                    continue
+                logger.info(
+                    "Rebuilding table %s to make foreign key constraints deferrable",
+                    table_name,
+                )
+                try:
+                    # Rebuilding the table from the current model state regenerates the column
+                    # definitions, which SQLite always emits with the deferrable clause (see
+                    # DatabaseSchemaEditor.sql_create_inline_fk). Only tested with Django 3.2
+                    schema_editor._remake_table(model)
```
rtibbles, line 124: I had never heard of this function before this week, now I have reviewed two PRs that contain it!

https://github.com/learningequality/kolibri/pull/14881/changes#diff-ca76c44748846b391bdb2cebc1b735de22710e703b7c79cb50d8bf6f2b70de14R30

rtibbles (approved):
Code and test changes make sense to me.

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,136 @@
...
+                except DatabaseError as e:
```
rtibblesbot, line 125: **suggestion:** `DatabaseError` is caught here but not re-raised. With `atomic=False` on the migration, there is no wrapping transaction to roll back, so Django will mark the migration as applied even if some table rebuilds silently failed — and the idempotency check at the top means those tables will never be retried. If the intent is best-effort (rebuild as many tables as possible rather than aborting on first failure), document that contract explicitly and emit a summary log at the end listing any tables that failed, so operators can identify and repair them manually. If full atomicity is desired, remove the `try/except` and let the exception propagate.

tests/testapp/tests/test_deferrable_foreign_keys.py
```diff
@@ -0,0 +1,87 @@
...
+        with connection.schema_editor(atomic=False) as schema_editor:
+            op(global_apps, schema_editor)
+
+        # now deferrable again, and data preserved
+        self.assertIn("DEFERRABLE", _get_table_sql(connection, table))
+        self.assertTrue(Facility.objects.filter(id=f_id).exists())
+
+    def test_iter_apps__all(self):
+        op = MakeForeignKeysDeferrable()
+        self.assertEqual([app.name for app in op._iter_apps(global_apps)], [
```
rtibblesbot, line 70: **suggestion:** This assertion hardcodes the full ordered `INSTALLED_APPS` list. Any addition, removal, or reordering of apps in the test settings will break this test for reasons unrelated to `_iter_apps`. Derive the expected value dynamically instead:
```python
expected = [app.name for app in global_apps.get_app_configs()]
self.assertEqual([app.name for app in op._iter_apps(global_apps)], expected)
```

rtibblesbot (commented):
All 6 prior findings resolved. Two new suggestions, one coverage gap.

CI is failing on `contributor-gate` due to an expired App token — infrastructure, not test failure.

- **suggestion** (inline): `deferrable_foreign_keys.py:125` — `DatabaseError` caught and not re-raised; with `atomic=False` on the migration, Django still records it as applied even if individual table rebuilds silently fail. If the intent is best-effort, document the contract explicitly and consider logging which tables were skipped so operators can identify them.
- **suggestion** (inline): `test_deferrable_foreign_keys.py:70` — `test_iter_apps__all` hardcodes the full `INSTALLED_APPS` list; any change to test settings breaks this test for unrelated reasons. Prefer `[app.name for app in global_apps.get_app_configs()]` as the expected value.
- **suggestion**: The three `test_iter_apps__*` tests only use apps where `label == name` (`'morango'`, `'facility_profile'`), so the `app_config.label` vs `app_config.name` fix has no regression coverage. A one-liner would lock it in:
  ```python
  def test_iter_apps__include_by_label(self):
      # 'auth' label != 'django.contrib.auth' name
      op = MakeForeignKeysDeferrable(include_app_labels=['auth'])
      self.assertEqual([app.name for app in op._iter_apps(global_apps)], ['django.contrib.auth'])
  ```

RESOLVED — morango/deferrable_foreign_keys.py — `app_config.name` vs `app_config.label` mismatch
RESOLVED — morango/deferrable_foreign_keys.py — `_remake_table` private-API documentation
RESOLVED — morango/deferrable_foreign_keys.py — mixed immediate/deferrable FK edge case
RESOLVED — morango/deferrable_foreign_keys.py — `atomic=True` in `run()` vs `atomic=False` in test
WITHDRAWN — tests/testapp/tests/test_deferrable_foreign_keys.py:70 — ordered list assertion (withdrawn prior round)
RESOLVED — tests/testapp/tests/test_deferrable_foreign_keys.py — string-formatting table name without `quote_name`

</details>

---

</details>

Issue #339, opened 2026-06-25
Connection exception during transfer does not always retry


bjester:
**Target branch:** release-v0.8.x 

## Context
Morango's sync session controller uses the `Retry` mechanism of python `requests`.

## Observed behavior

If connection interruptions cause a malformed response, errors like `ProtocolError` can cause the sync transfer to all out fail. This could occur due to low or flaky connectivity. 

## Errors and logs

```
ERROR 2026-06-09 11:38:18,722 morango.sync.session ChunkedEncodingError Reason: (no response)
ERROR 2026-06-09 11:38:18,722 morango.sync.controller ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read))
INFO 2026-06-09 11:38:18,737 kolibri.core.auth.management.utils Receiving data (14600/111908, 54.25MB)
ERROR 2026-06-09 11:38:18,779 kolibri.core.tasks.job Job fd71f20d5dee4747bb22320c39c1f4b3 raised an exception: Traceback (most recent call last):
  File "C:\Python38\Lib\site-packages\kolibri\dist\urllib3\response.py", line 761, in _update_chunk_length
    self.chunk_left = int(line, 16)
ValueError: invalid literal for int() with base 16: b''

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python38\Lib\site-packages\kolibri\dist\urllib3\response.py", line 444, in _error_catcher
    yield
  File "C:\Python38\Lib\site-packages\kolibri\dist\urllib3\response.py", line 828, in read_chunked
    self._update_chunk_length()
  File "C:\Python38\Lib\site-packages\kolibri\dist\urllib3\response.py", line 765, in _update_chunk_length
    raise InvalidChunkLength(self, line)
urllib3.exceptions.InvalidChunkLength: InvalidChunkLength(got length b'', 0 bytes read)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python38\Lib\site-packages\kolibri\dist\requests\models.py", line 760, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "C:\Python38\Lib\site-packages\kolibri\dist\urllib3\response.py", line 624, in stream
    for line in self.read_chunked(amt, decode_content=decode_content):
  File "C:\Python38\Lib\site-packages\kolibri\dist\urllib3\response.py", line 857, in read_chunked
[...]

PR #342, opened 2026-06-30
Additional retry protection throughout sync pipeline

bjester:
## Summary
- Updates `urllib3` imports to import directly from the package instead of through `requests.packages` proxy, because it stopped bundling the dependency nearly a decade ago. Dependency restrictions were added for `requests` and `urllib3` to enforce compatible versions for Morango's use. Minimum `requests` version matches pinned version in Kolibri 0.19.x
- Updates buffer creation during transfer to be idempotent-- the same buffer chunk can be pulled or pushed multiple times without issue or inflating record transfer count
- Makes the sync session requests to close sync or transfer sessions passive to 404s. A 404 would likely indicate it has already happened due to the `active=True` filter.
- Adds retry behavior to the `SessionWrapper` that utilizes the `Retry` utility, if configured, to retry low-level connection issues that are not automatically retried by `urllib3`.
- Refactors bandwidth tracking in the `SessionWrapper` to be more self-contained

## TODO

- [X] Has changelog been written/updated?

## Reviewer guidance
Morango integration tests in Kolibri are passing locally with these changes. 

Careful attention was given to version specific gotchas and to ensure support for Kolibri's supported python versions. This PR also focuses on lower-level retries-- retrying complex behaviors like certificate requests, which would require new nonces, are not addressed. The transfer session API is passive to one already existing, to support resumption, as long as the request is the same. Overall, the most important area for retries is the buffer transfer since it would involve many requests, increasing the likelihood of a failure.

I've added comments for the reasoning of specific changes.

## Issues addressed

Closes https://github.com/learningequality/morango/issues/339

## AI Usage
AI was used to jumpstart the changes, although its approach got convoluted quickly. So most of this is handcrafted to integrate with `requests` and `urllib3` as smoothly as possible. AI was used to create some tests and to keep tests up-to-date through several iterations of this work.


morango/sync/syncsession.py
```diff
@@ -76,6 +78,24 @@ def _get_client_ip_for_server(server_host, server_port):
...
+    """
+    Decorator that wraps callables to ignore 404s caused by its use of requests
+    :param target: A callable
+    :return: A callable
+    """
+    def wrapper(*args, **kwargs):
+        try:
+            return target(*args, **kwargs)
+        except HTTPError as e:
+            if not e.response or e.response.status_code != 404:
```
rtibblesbot: **blocking:** This never ignores a real 404. `requests.exceptions.HTTPError` raised by `Response.raise_for_status()` carries `response=self` (the errored response), and `requests.Response.__bool__` returns `self.ok` — `False` for any `status_code >= 400`. So `not e.response` is `True` for *any* HTTPError reaching this decorator, the `or` short-circuits, and `raise e` fires unconditionally regardless of status code:
```python
>>> r = requests.Response(); r.status_code = 404
>>> not HTTPError(response=r).response
True
```
The new tests pass only because they raise `HTTPError(response=mock.Mock())` — a generic `Mock` is truthy regardless of the `status_code` set on it, so the test exercises a branch real responses never take. Fix: `if e.response is None or e.response.status_code != 404:`.

morango/sync/session.py
```diff
@@ -45,59 +101,129 @@ def __init__(self):
...
+                    response = e.response
+
+                # Retry for transport-level failures where we did not get a usable response body.
+                if (
+                    retries is not None
+                    and isinstance(e, _RETRY_REQUEST_EXCEPTIONS)
+                    and _is_retryable_method(retries, method)
+                ):
+                    # may raise if retries have been exhausted
+                    retries = retries.increment(
```
rtibblesbot: **blocking:** `Retry.increment()` raises `urllib3.exceptions.MaxRetryError` directly when retries are exhausted (confirmed in the installed urllib3's `Retry.increment` source) — and `MaxRetryError` is not a subclass of `requests.exceptions.RequestException`. Since `increment()` is called here rather than inside `HTTPAdapter.send()` (which normally translates it), the exception propagates straight out of `request()` as a bare `urllib3` exception. That:
- skips `_log_response_error(e, response)` below for exactly the failure that matters most
- breaks the docstring's contract that this method "raises requests.exceptions.RequestException"
- means `ignore_404`'s `except HTTPError` (and any other `requests.exceptions` catch in the codebase) won't catch it

Consider catching `urllib3.exceptions.MaxRetryError` here and re-raising as a `requests.exceptions` type (or logging before letting it propagate).

setup.py
```diff
@@ -28,7 +28,7 @@
         "rsa<4.10",
         "djangorestframework>3.10",
         "django-ipware==4.0.2",
-        "requests",
+        "requests>2.17",
```
rtibblesbot: **suggestion:** `session.py`/`syncsession.py` now import `urllib3.util.retry.Retry` and `urllib3.util.url.parse_url` directly (previously via `requests.packages.urllib3`), but `urllib3` isn't declared in `install_requires` — it's only available transitively through `requests`. Given the wide `requests>2.17` range and the `_is_retryable_method` shim for the `method_whitelist`→`allowed_methods` rename, worth declaring `urllib3` as a direct dependency with an explicit range.

morango/sync/session.py
```diff
@@ -45,59 +101,129 @@ def __init__(self):
             if instances:
                 user_agent_header += " " + "{}/{}".format(instances[0], SETTINGS.CUSTOM_INSTANCE_INFO.get(instances[0]))
         self.headers["User-Agent"] = "{} {}".format(user_agent_header, self.headers["User-Agent"])
+        self.hooks["response"].append(self._track_bytes_received)
```
rtibblesbot, line 128: **suggestion:** This hook now fires on every response the underlying `Session` produces, including attempts that get retried and intermediate redirect hops — but `bytes_sent` accounting in `_track_bytes_sent` isn't tracked with the same symmetry. Since these counters are persisted to `TransferSession.bytes_sent`/`bytes_received` and surfaced via the API, a connection that needed several retries will now report inflated transfer stats. Worth deciding whether retried-attempt bytes should count at all.

morango/sync/session.py
```diff
@@ -24,19 +39,60 @@ def _headers_content_length(headers):
...
+    :type retries: urllib3.util.retry.Retry
+    :type method: str
+    :rtype: bool
+    """
+    allowed_methods = getattr(retries, "allowed_methods", None)
+    if allowed_methods is None:
+        allowed_methods = getattr(retries, "method_whitelist", None)
+    if allowed_methods is False or allowed_methods is None:
+        return True
+    return method.upper() in allowed_methods
```
rtibblesbot: **suggestion:** The adapter's `Retry` (syncsession.py:156) is constructed without `allowed_methods`, so it defaults to urllib3's `{'HEAD','GET','PUT','DELETE','OPTIONS','TRACE'}` — POST is excluded. `_is_retryable_method` here will therefore never allow retrying buffer-push POSTs, even though the companion `utils.py` change makes push processing idempotent specifically to tolerate retries. Worth confirming whether `allowed_methods` should include `"POST"` for this adapter.

morango/sync/syncsession.py
```diff
@@ -76,6 +78,24 @@ def _get_client_ip_for_server(server_host, server_port):
     return IP
 
 
+def ignore_404(target):
+    """
+    Decorator that wraps callables to ignore 404s caused by its use of requests
+    :param target: A callable
+    :return: A callable
+    """
+    def wrapper(*args, **kwargs):
```
rtibblesbot, line 88: **nitpick:** `wrapper` doesn't use `functools.wraps(target)`, so `_close_transfer_session`/`_close_sync_session` lose `__name__`/`__doc__`/`__wrapped__` for introspection. Harmless today since the debug log reads `target.__name__` from the closure, but worth the one-line fix.

morango/sync/utils.py
```diff
@@ -128,18 +128,26 @@ def validate_and_create_buffer_data(  # noqa: C901
...
             transfer_session.bytes_sent = connection.bytes_sent
         if connection is not None:
             transfer_session.bytes_received = connection.bytes_received
 
-        transfer_session.save()
-
         Buffer.objects.bulk_create(buffer_list)
         RecordMaxCounterBuffer.objects.bulk_create(rmcb_list)
 
+        transfer_session.records_transferred = Buffer.objects.filter(transfer_session=transfer_session).count()
```
rtibblesbot: **praise:** Deleting existing `Buffer`/`RecordMaxCounterBuffer` rows by `model_uuid` before `bulk_create`, and recomputing `records_transferred` from a `count()` instead of `+=`, is exactly right given the `unique_together` constraint — fixes the retry-duplicate problem without overreaching.

rtibblesbot (changes requested):
Two passes converge on the same root issue from different angles: the new retry/cleanup machinery in `session.py`/`syncsession.py` doesn't fully honor the `requests` exception boundary it relies on. Verified both findings directly against the installed `requests`/`urllib3` source below. CI: one unrelated bot-permission check failing, all Python test matrices passing/pending.

**Blocking**
- `ignore_404` (syncsession.py:91) never ignores a real 404 — `Response.__bool__` returns `self.ok`, so `not e.response` is `True` for any errored response and the `raise e` branch fires unconditionally. The tests pass only because they use a generic `Mock()`, which is truthy regardless of `status_code`.
- Exhausted retries (session.py:214) raise `urllib3.exceptions.MaxRetryError` directly, bypassing the `requests.exceptions.RequestException` boundary the method's own docstring promises. This skips `_log_response_error` for the final failure and means `ignore_404`'s `except HTTPError` won't catch it either.

**Suggestions**
- session.py:104 — bytes tracking is no longer 1:1 with logical requests: the `response` hook fires on every retry attempt and redirect hop, but `bytes_sent` accounting doesn't track retries symmetrically. These counters are persisted to `TransferSession` and exposed via the API, so a flaky connection now reports inflated transfer stats.
- setup.py:31 — `urllib3.util.retry.Retry`/`urllib3.util.url.parse_url` are now imported directly rather than via `requests.packages.urllib3`, but `urllib3` isn't declared in `install_requires`.
- session.py:69 — the adapter's `Retry` (syncsession.py:156) uses urllib3's default `allowed_methods`, which excludes POST, so buffer pushes don't benefit from this new transport-level retry even though the companion `utils.py` change makes push processing idempotent specifically to tolerate retries.

**Nitpick**
- `ignore_404` (syncsession.py:87) doesn't use `functools.wraps`, so decorated methods lose `__name__`/`__doc__`.

---

</details>

morango/sync/syncsession.py
bjester: Fixed

morango/sync/session.py
bjester: Fixed

setup.py
bjester: Fixed

morango/sync/session.py
bjester: Made the tracking as close as possible without getting deep into urllib3.

morango/sync/session.py
bjester: Made it allow any method

morango/sync/syncsession.py
rtibblesbot: Confirmed — `e.response is None or e.response.status_code != 404` correctly handles the `Response.__bool__` gotcha now. Thanks for the fix.

morango/sync/syncsession.py
bjester: Fixed

morango/sync/session.py
rtibblesbot: Confirmed — `MaxRetryError` is now caught in `_should_retry` and re-raised as `requests.exceptions.RetryError`, so it stays within the `RequestException` contract. Thanks for the fix.

setup.py
rtibblesbot: Confirmed — `urllib3>=1.21.1` is now declared directly in `install_requires`. Thanks for the fix.

setup.py
```diff
@@ -28,7 +28,8 @@
...
+        "urllib3>=1.21.1",
```
rtibblesbot: **blocking:** `urllib3>=1.21.1` doesn't guarantee the `Retry.allowed_methods` kwarg.
- That kwarg was only added in urllib3 1.26.0 (verified directly: 1.25.11's `Retry.__init__` only accepts `method_whitelist`).
- `NetworkSyncConnection.__init__` (syncsession.py:158-162) passes `allowed_methods=None` unconditionally.
- On urllib3 1.21.1–1.25.x, which this floor permits, constructing a connection raises `TypeError: __init__() got an unexpected keyword argument 'allowed_methods'`, breaking sync entirely.
- Fix: bump the floor to `urllib3>=1.26.0`, or build the kwarg dynamically the way `_is_retryable_method` already does on the read side (checks `allowed_methods`, falls back to `method_whitelist`).

morango/sync/syncsession.py
```diff
@@ -133,7 +155,11 @@ def __init__(
         self.session = SessionWrapper()
         # sleep for {backoff factor} * (2 ^ ({number of total retries} - 1)) between requests
         # with 7 retry attempts, sleep escalation becomes (0.6s, 1.2s, ..., 38.4s)
-        retry = Retry(total=retries, backoff_factor=backoff_factor)
+        retry = Retry(
+            total=retries,
+            backoff_factor=backoff_factor,
+            allowed_methods=None,  # allow any method
```
rtibblesbot: **suggestion:** `allowed_methods=None` plus `_is_retryable_method`'s None-is-retryable logic now makes POST retryable on `ConnectionError`/`ChunkedEncodingError`.
- Failure mode: `_create_sync_session` and `_push_certificate_chain` consume a single-use nonce server-side (`Nonce.use_nonce`). If the transient failure happens after the server already processed the request (e.g. while reading the response), the resend hits `403 Nonce is not valid` instead of the original transient error.
- Coverage gap: no test exercises retry against a nonce-consuming POST — the only retry test (`test_get_remote_certs__retries_chunked_encoding_error`) is a GET.

morango/sync/session.py
```diff
@@ -45,58 +102,170 @@ def __init__(self):
...
+        :rtype: requests.Response
+        :raises requests.exceptions.RequestException: If a request-related error occurs and the
+            request cannot be retried or ultimately fails after retries.
+        """
+        response = None
+        try:
+            response = super(SessionWrapper, self).request(method, url, **kwargs)
+            response.raise_for_status()
+            return response
+        except Exception as e:
```
rtibblesbot: **suggestion:** `request()` now catches bare `Exception` instead of `exceptions.RequestException` (its own docstring still says the latter). Harmless today since it re-raises unchanged, but it will funnel unrelated bugs (e.g. a `TypeError` from bad kwargs) through the same "request failed" log line, obscuring the real failure class during debugging. Consider narrowing back to `RequestException`.

morango/sync/session.py
```diff
@@ -45,58 +102,170 @@ def __init__(self):
...
+            self._track_bytes_sent(request)
+            try:
+                response = super(SessionWrapper, self).send(request, **kwargs)
+                return response
+            except exceptions.RequestException as e:
+                retries = self._should_retry(retries, request, e, response=response)
+                if retries is None:
+                    raise e
+
+    def _should_retry(self, retries, request, e, response=None):
```
rtibblesbot, line 273: **suggestion:** `_should_retry` is exactly the code that fixed last round's `MaxRetryError`→`RequestException` boundary finding, but no test drives it to exhaustion. Existing coverage (`test_get_remote_certs__retries_chunked_encoding_error`) only exercises one successful retry on a GET. A test that exhausts `total` retries and asserts `requests.exceptions.RetryError` propagates (not a bare `MaxRetryError`) would lock in the fix rather than relying on code reading alone.

rtibblesbot (changes requested):
6 of 6 prior actionable findings resolved (1 acknowledged, unchanged); 1 new blocking finding plus a few suggestions from this pass (see inline).

**CI:**
- Contributor-check failure is infra/permissions, not code (consistent with prior round).
- Python test matrix was still pending at review time.

**New findings:**
- blocking: dependency floor mismatch — see inline on `setup.py:32`
- suggestion: method-agnostic retry can resend a single-use-nonce POST — see inline on `morango/sync/syncsession.py:161`
- suggestion: broadened exception catch in `request()` — see inline on `morango/sync/session.py:197`
- suggestion: retry-exhaustion path has no dedicated test — see inline on `morango/sync/session.py:227`

Praise: moving the retry loop into an overridden `send()` (session.py) is a cleaner fix than the old self-invoking `request()` shim, and is what let the `MaxRetryError`→`RetryError` translation land without duplicating retry-decision logic.

RESOLVED — morango/sync/syncsession.py:93 — `ignore_404` never ignored real 404s (`Response.__bool__` gotcha)
RESOLVED — morango/sync/session.py:214 — `MaxRetryError` bypassed the `requests.exceptions.RequestException` boundary
RESOLVED — setup.py:31 — `urllib3` imported directly but not declared in `install_requires`
RESOLVED — morango/sync/session.py:104 — `bytes_sent` tracking not symmetric with retries
RESOLVED — morango/sync/session.py:69 — adapter `Retry` excluded POST from retryable methods
RESOLVED — morango/sync/syncsession.py:87 — `ignore_404` wrapper missing `functools.wraps`
ACKNOWLEDGED — morango/sync/utils.py:148 — praise on idempotent buffer replace-then-recount (still valid, unchanged)

</details>

---

</details>

morango/sync/session.py
bjester: How exactly does this obscure the failure class?

rtibblesbot: You're right — I missed that `_log_response_error` logs `err.__class__.__name__` explicitly (session.py:88), so the actual exception class is printed in the log line regardless of which type is caught. There's no obscuring. Withdrawing this finding.

setup.py
bjester: Fixed

morango/sync/syncsession.py
bjester: Dealing with multiple orchestrated requests and responses requires a whole other level of behavior which this doesn't target. I'm not going to implement that in this PR.

setup.py
rtibblesbot: Confirmed — `urllib3>=1.26.0` now guarantees the `allowed_methods` kwarg. Thanks for the fix.

morango/sync/syncsession.py
rtibblesbot: Understood — that's a reasonable scope boundary for this PR. Not pushing further on it.

rtibblesbot (commented):
2 of 4 prior findings resolved; 1 still open, 1 acknowledged (see collapsed detail below).

- **suggestion (unaddressed):** `allowed_methods=None` still lets the shared session retry `_create_sync_session`'s nonce-protected POST — see inline on `morango/sync/syncsession.py:161`.
- **suggestion (new):** per-chunk `Buffer.objects.filter(...).count()` turns O(1) counting into an O(n) query every chunk — see inline on `morango/sync/utils.py:148`.
- **nitpick (new):** `_close_transfer_session` lacks the 500-raises test its `_close_sync_session` sibling has — see inline on `morango/sync/syncsession.py:507`.

CI: same unrelated contributor-check infra failure as prior rounds; all Python test matrices passing or still pending.

RESOLVED — setup.py:32 — urllib3 floor not declared alongside `allowed_methods` usage
UNADDRESSED — morango/sync/syncsession.py:161 — method-agnostic retry can resend a single-use-nonce POST
ACKNOWLEDGED — morango/sync/session.py:193 — broadened `except Exception` in `request()` (docstring updated to match)
RESOLVED — morango/sync/session.py:227 — retry-exhaustion path now has a dedicated test (`test_should_retry__raises_retry_error_on_max_retries`)
ACKNOWLEDGED — morango/sync/utils.py:148 — idempotent buffer replace-then-recount (unchanged)

</details>

---

</details>

morango/sync/syncsession.py
rtibblesbot: **suggestion:** Confirmed via `morango/api/viewsets.py` `SyncSessionViewSet.create` (~line 289): `Nonce.use_nonce` consumes the nonce before the sync session is created and the response returned. With `allowed_methods=None`, this shared session now retries `_create_sync_session`'s POST on `ConnectionError`/`ChunkedEncodingError` — errors that fire *after* the server already processed the request. A retry resends the same, now-consumed nonce and gets back `"Nonce is not valid"` (403) instead of the session that was actually created. Worth scoping which POST endpoints are safe to retry (or special-casing the nonce-invalid response) rather than enabling retries for every POST on this connection.

morango/sync/utils.py
rtibblesbot: **suggestion:** This recomputes the full `Buffer` count from scratch on every chunk (O(n) per chunk, ~O(n²/chunk_size) overall), replacing what was an O(1) increment. Fixes real double-counting on retried chunks so it's likely an acceptable trade-off, but worth confirming it isn't a bottleneck on large full-resync pushes.

morango/sync/syncsession.py
```diff
@@ -478,13 +504,15 @@ def _update_transfer_session(self, data, transfer_session):
             json=data,
         )
 
+    @ignore_404
```
rtibblesbot, line 504: **nitpick:** `_close_sync_session` has both an ignore-404 test and a raises-500 test; `_close_transfer_session` only has the 404 case. Same `ignore_404` decorator so risk is low, but a `test_close_transfer_session_raises_500` would keep coverage symmetric.

bjester: We don't need symmetric coverage. The purpose is a test for the ignore_404 behavior on a real method. Coverage is good. Adding more tests doesn't increase coverage.

morango/sync/utils.py
bjester: Updated to leverage the deleted count to use the diff to increment the count.

morango/sync/syncsession.py
rtibblesbot: Fair point — the ignore_404 behavior is already covered on a real method, and a second test wouldn't add meaningful coverage. Withdrawing this nitpick.

morango/sync/syncsession.py
bjester: Added `is_retryable` flag for explicit allowance of retries

morango/sync/session.py
```diff
@@ -1,19 +1,38 @@
...
+from morango.utils import nullcontext
 from morango.utils import serialize_capabilities_to_client_request
 from morango.utils import SETTINGS
 
 
 logger = logging.getLogger(__name__)
 
 
+_RETRY_REQUEST_EXCEPTIONS = (
+    exceptions.ConnectionError,
```
rtibblesbot, line 21: **blocking:**
- `ConnectionError` is exactly what `requests.adapters.HTTPAdapter.send()` raises after urllib3's own `conn.urlopen(..., retries=self.max_retries, ...)` has already exhausted its retry budget (the generic `except MaxRetryError` fallback in `requests/adapters.py` raises `ConnectionError(e, request=request)`).
- Including it here means `SessionWrapper.send()`'s outer loop retries the entire request again with a fresh `Retry` budget each time — up to `outer_retries × inner_retries` attempts (e.g. 7×7=49) before giving up, each with its own escalating backoff.
- That can turn the intended bounded retry policy into a multi-minute-to-tens-of-minutes hang during a real outage, working against this PR's own goal of bounding retries.
- `ChunkedEncodingError`/`ContentDecodingError` don't have this problem — they're raised later while reading/decoding the response body, a path urllib3's own retry never touches.
- Suggest dropping `ConnectionError` from this tuple, or narrowing to a subtype that bypasses urllib3's retry if one exists.

morango/sync/session.py
```diff
@@ -24,79 +43,275 @@ def _headers_content_length(headers):
...
+        :param response: Optional. The HTTP response object corresponding to the request, if available.
+        :type response: Optional[requests.Response]
+        :return: The new Retry object if the retry is allowed, otherwise None.
+        :rtype: Optional[urllib3.util.retry.Retry]
+        """
+        if response is None:
+            response = e.response
+
+        if (
+            retries is None
```
rtibblesbot, line 295: **suggestion:** This gate only short-circuits `is_retryable=False` calls because POST happens to fall outside urllib3's default `allowed_methods` set — `retries` here is the adapter's default zero-budget `Retry(0)`, not `None`. A future GET/PUT/DELETE call site that forgets `is_retryable=True` would fall through to `retries.increment(...)`, immediately hit `MaxRetryError` (total=0), and have its real exception (e.g. `ConnectionError`) silently replaced by `RetryError` with zero actual retry attempts. Consider passing `retries=None` explicitly through the non-opted-in path (the `nullcontext` branch in `request()`) so this always short-circuits via `retries is None` regardless of HTTP method.

morango/utils.py
```diff
@@ -143,3 +143,16 @@ def exception_path(exc):
...
+
+
+class nullcontext:
+    """Replace this with contextlib.nullcontext when python3.6 support is dropped"""
+
+    def __init__(self, value=None):
+        self.value = value
+
+    def __enter__(self, *args, **kwargs):
+        return self.value or self
```
rtibblesbot: **suggestion:** `self.value or self` diverges from `contextlib.nullcontext`, which this class is documented as a stand-in for: any falsy `value` (`0`, `False`, `""`) causes `__enter__` to return the `nullcontext` instance itself instead of `value`. Dormant today (only called as `nullcontext()` with no value bound via `with ... as`), but as a general-purpose utility this is a correctness gap. Suggest `return self.value` to match stdlib semantics.

morango/sync/session.py
```diff
@@ -24,79 +43,275 @@ def _headers_content_length(headers):
...
 class SessionWrapper(Session):
     """
     Wrapper around `requests.sessions.Session` in order to implement logging around all request errors.
     """
 
-    bytes_sent = 0
-    bytes_received = 0
-
-    def __init__(self):
+    def __init__(self, max_retries):
```
rtibblesbot, line 114: **nitpick:** `SessionWrapper(max_retries)` now requires the arg with no default, unlike the old no-arg constructor — acceptable since this class isn't part of morango's public API.

rtibblesbot (changes requested):
Delta review of `c6611a06` — same SHA as my last COMMENTED verdict, no code changed. Re-verified the diff from scratch and found one new blocking issue in the retry loop plus two dormant-but-worth-fixing correctness gaps.

2 of 2 prior findings remain ACKNOWLEDGED (deferred, code unchanged) — see collapsed block below.

New findings (see inline comments):
- **blocking** — `ConnectionError` is retried by the new outer loop even though it's exactly the exception urllib3 raises *after already exhausting its own retry budget* — compounds to up to N² connection attempts and can turn a bounded retry policy into a multi-minute hang.
- **suggestion** — the retry gate in `_should_retry` only no-ops for non-opted-in calls because POST happens to fall outside urllib3's default allowed-methods set; a future GET/DELETE call site that forgets `is_retryable=True` would get its real exception silently replaced by `RetryError`.
- **suggestion** — `nullcontext.__enter__` returns `self` instead of `value` for falsy values, diverging from `contextlib.nullcontext`. Dormant today.
- **nitpick** — `SessionWrapper.__init__` now requires `max_retries` positionally (fine, just flagging for any external callers).

CI:
- One failing check is unrelated org-workflow infra (missing secrets in `learningequality/.github`'s shared workflow).
- A few Python-version jobs were still pending at review time.
- Everything else passes.

ACKNOWLEDGED — morango/sync/session.py:240 — `request()` catches bare `Exception` instead of `exceptions.RequestException`
ACKNOWLEDGED — morango/sync/syncsession.py:504 — `_close_transfer_session` missing a `raises_500` test symmetric to `_close_sync_session`

</details>

---

</details>

morango/sync/session.py
bjester: Checked if `retries.total == 0`

morango/sync/session.py
bjester: Added check for wrapped `MaxRetryError`

morango/utils.py
bjester: Fixed

rtibblesbot: Confirmed — `__enter__` now returns `self.value` directly, matching `contextlib.nullcontext` semantics for falsy values too. Thanks for the fix.

morango/sync/session.py
rtibblesbot: Confirmed — the `any(isinstance(arg, MaxRetryError) for arg in e.args)` check catches the wrapped case (`requests` raises `ConnectionError(e, request=request)` with the `MaxRetryError` as `e.args[0]`), so `_should_retry` now returns `None` and the outer loop no longer restarts a fresh retry budget on top of an already-exhausted one. Thanks for the fix.

morango/sync/session.py
rtibblesbot: Confirmed — checking `retries.total == 0` short-circuits regardless of HTTP method, so this no longer depends on POST happening to fall outside `allowed_methods`. Thanks for the fix.

morango/sync/syncsession.py
```diff
@@ -461,31 +484,33 @@ def _create_sync_session(self, data):
...
         )
 
     def _update_transfer_session(self, data, transfer_session):
         return self.session.patch(
             self.urlresolve(api_urls.TRANSFERSESSION, lookup=transfer_session.id),
-            json=data,
+            json=data, is_retryable=True
         )
 
+    @ignore_404
```
rtibblesbot, line 504: **suggestion:** (unaddressed from prior review) `_close_sync_session` has `test_close_sync_session_raises_500` verifying non-404 errors propagate through `ignore_404`, but `_close_transfer_session` has no equivalent `test_close_transfer_session_raises_500`. Since both are wrapped by the same decorator, a symmetric test would close the coverage gap.

morango/sync/session.py
```diff
@@ -24,79 +43,280 @@ def _headers_content_length(headers):
...
+        else:
+            ctx = nullcontext()
+
+        with ctx:
+            response = None
+            try:
+                response = super(SessionWrapper, self).request(method, url, **kwargs)
+                response.raise_for_status()
+                return response
+            except Exception as e:
```
rtibblesbot, line 240: **suggestion:** `request()` now catches bare `Exception` instead of the previous `exceptions.RequestException`. This means a programming error inside `Session.request()` (e.g. a bad kwarg `TypeError`, a bug in a response hook) gets funneled through `_log_response_error` and logged as a request/HTTP error, misdirecting debugging. Consider scoping this back to `exceptions.RequestException`, which is the actual contract this method handles.

morango/sync/utils.py
```diff
@@ -128,18 +128,26 @@ def validate_and_create_buffer_data(  # noqa: C901
         buffer_list += [Buffer(**record)]
 
     with transaction.atomic():
-        transfer_session.records_transferred += len(data)
+        deleted_buffers, _ = Buffer.objects.filter(
```
rtibblesbot, line 131: **praise:** Deleting existing `Buffer`/`RecordMaxCounterBuffer` rows by `model_uuid` before `bulk_create`, then netting `records_transferred` against `deleted_buffers`, makes the newly-retryable push/pull chunk POST idempotent under retry — covered by `test_push_repeat_chunk`/`test_pull_repeat`.

rtibblesbot (commented):
Re-review at the same HEAD SHA (`ddb551fc`) as the prior COMMENTED verdict. 1 of 2 prior findings resolved; 1 still open (see below). One new suggestion from this pass. CI: all real test suites pass; the only failing check (`Check if author is contributor`) is unrelated repo automation.

- **suggestion** — `syncsession.py:504` — still missing a `test_close_transfer_session_raises_500` symmetric to `test_close_sync_session_raises_500` (see inline).
- **suggestion** — `session.py:240` — `request()` now catches bare `Exception` instead of `exceptions.RequestException` (see inline).

RESOLVED — morango/sync/session.py:222 — `request()` docstring vs bare-Exception catch mismatch
UNADDRESSED — morango/sync/syncsession.py:504 — missing `test_close_transfer_session_raises_500`

</details>

---

</details>

morango/sync/syncsession.py
bjester: I already told you I wasn't going to add that because it doesn't really add any meaningful coverage

morango/sync/session.py
bjester: The error gets logged with the response output, if any, because that gets useful, but it never obscures the error. This is a non-issue.

rtibblesbot: You're right — I re-raised this without re-checking; `raise e` preserves the original exception type/traceback and `_log_response_error` logs `err.__class__.__name__`, so a programming error still surfaces as itself. Withdrawing this finding (and apologies, I already withdrew the same point in the prior review round and shouldn't have re-raised it here).

morango/sync/syncsession.py
rtibblesbot: Understood, dropping this one.

morango/sync/operations.py
```diff
@@ -1390,9 +1390,8 @@ def close_transfer_session(self, context):
         Closes remote transfer session
 
         :type context: NetworkSessionContext
-        :return: The Response
         """
-        return context.connection._close_transfer_session(context.transfer_session)
+        context.connection._close_transfer_session(context.transfer_session)
```
bjester, line 1394: The response object wasn't really necessary, so this no longer returns it.

morango/sync/operations.py
```diff
@@ -1763,8 +1762,5 @@ def handle(self, context):
         """
         :type context: NetworkSessionContext
         """
-        response = self.close_transfer_session(context)
-        remote_status = transfer_statuses.COMPLETED
-        if response.status_code < 200 or response.status_code >= 300:
-            remote_status = transfer_statuses.ERRORED
-        return remote_status
+        self.close_transfer_session(context)
+        return transfer_statuses.COMPLETED
```
bjester, line 1766: Before this code, we call `raise_for_status()` so this seemed unnecessary and because we may get a 404 with a retry, we wouldn't have an accurate response. This now relies on the controller to set the status.

tests/testapp/tests/test_api.py
```diff
@@ -1025,11 +1045,9 @@ def make_buffer_get_request(
                 transfer_session_id=t_id, model_uuid__in=model_uuids
             ).delete()
 
-            # run the validation logic to ensure no errors were returned
-            errors = validate_and_create_buffer_data(
```
bjester, line 1029: This function didn't actually return anything.

setup.py
```diff
@@ -28,7 +28,8 @@
         "rsa<4.10",
         "djangorestframework>3.10",
         "django-ipware==4.0.2",
-        "requests",
+        "requests>=2.27.1",
+        "urllib3>=1.26.0",
```
bjester, line 32: Needed to pin above this because of an API change to `Retry`

morango/sync/utils.py
```diff
@@ -128,18 +128,26 @@ def validate_and_create_buffer_data(  # noqa: C901
...
+            transfer_session=transfer_session,
+            model_uuid__in=[record["model_uuid"] for record in data]
```
bjester, line 133: Default chunk size is `500`. I think this should be okay?

morango/sync/session.py
```diff
@@ -24,79 +43,280 @@ def _headers_content_length(headers):
...
+        ):
+            return None
+
+        # requests might wrap `MaxRetryError` in a `ConnectionError`
+        if any(isinstance(arg, MaxRetryError) for arg in e.args):
+            return None
+
+        try:
+            # may raise if retries have been exhausted
+            retries = retries.increment(
```
bjester, line 308: `increment` returns a new `Retry` object every time, hence why we pass it around.

morango/sync/utils.py
rtibblesbot: Confirmed — the delete/filter queries are scoped to `model_uuid__in=[record["model_uuid"] for record in data]`, i.e. bounded by the chunk itself, not the full transfer session's accumulated buffer set. So each chunk's DB work is O(chunk_size) regardless of how large the overall transfer gets — the O(n²) risk I flagged is gone independent of what the chunk size default is. Thanks for the fix.

morango/sync/controller.py
```diff
@@ -270,7 +270,7 @@ def _invoke_middleware(self, context, middleware):
             return context.stage_status
         except Exception as e:
             # always log the error itself
-            logger.error(e)
+            logger.exception(e)
```
rtibbles, line 273: 👍

morango/sync/session.py
```diff
@@ -24,79 +43,280 @@ def _headers_content_length(headers):
...
+    try:
+        response_content = response.content if response else "(no response)"
+    except Exception:
+        response_content = "(unable to read response)"
+    logger.error(
+        "{} Reason: {}".format(err.__class__.__name__, response_content)
+    )
+
+
+class ContextualRetryHTTPAdapter(HTTPAdapter):
```
rtibbles, line 92: Nice - I think I came across this recently somewhere either in Kolibri or ricecooker and had wondered why we hadn't used this approach before. I approve!

rtibbles (approved):
I understand the changes made here - I get the intention, and also it seems that they are all targeted at the same goal as the issue. The proof here will not come from my code review, so nothing else to say but approve!

Issue #344, opened 2026-07-07
Poor sync transfer performance due to buffer serialization


bjester:
**Target branch:** release-v0.8.x

## Current behavior

On a very large database, syncing 363k buffer records, and 728k related record max counters, took a significant amount of time to transfer data to another local instance. On average, a transfer of just 200 buffers took 10 seconds. During each buffer chunk, significant CPU usage was observed. Analyzing the process, it was observed that the code involved queries related to buffer serialization.

<img width="1200" height="730" alt="Image" src="https://github.com/user-attachments/assets/646d181a-35dc-4290-a309-ce32d9d83320" />

Upon investigation, the `BufferSerializer` creates an N+1 query problem when serializing a list of buffers. This is because the serialization of each buffer accesses `rmcb_list` on the buffer, creating a query for the counters, meaning 200 buffers creates 201 queries.

## Desired behavior

The queries for the `RecordMaxCounterBuffer` records should be consolidated into on query that can operate in bulk, to minimize the overhead of numerous queries to the DB.

## Value add

Speeds up transfer when syncing an instance with a large SQLite DB.

## Possible tradeoffs

We've observed this issue in use of DRF within Kolibri and Kolibri Studio. We've developed architectural patterns in those projects that help navigate these issues, but those are not available to Morango at this time. For the time being, a targeted fix is likely the best approach.

## AI usage
AI was used to verify the issue that was identified through manual investigation.

bjester:
Closed by https://github.com/learningequality/morango/pull/345

PR #345, opened 2026-07-07
Improve buffer serialization performance by performing bulk counter lookup

bjester:
## Summary

Improves the performance of sync transfer for a large database by performing a single query for counter records (for associated buffers being transferred). For the same DB referenced in the issue, where a chunk would take approximately 10 seconds to transfer, the same chunk takes less than 1 second-- a 90%+ decrease in the processing time.

Tests already cover the API and serializer integration with multiple buffers.

## Reviewer guidance

Is the existing test coverage sufficient or should unit tests be added for the serializer?

## Issues addressed

Closes https://github.com/learningequality/morango/issues/344

## AI Usage
AI was used to verify the performance issue, and with that context, implement the fix. The fix was minimal and existing tests covered the modified code. Small changes were made to the implementation to loop over transfer sessions, to make the code a little cleaner, even though it's only ever called with one transfer session.


morango/api/serializers.py
```diff
@@ -158,11 +160,38 @@ class Meta:
...
+            # Morango implementation will only ever call this for one transfer session ID at a time,
+            # but a loop makes the code straightforward regardless and limits the quantity of
+            # SQLite variables in use
+            for transfer_session_id in transfer_session_ids:
+                # bulk-fetch all RMCB records needed for this batch of buffers in a
+                # single query, then cache the relevant subset on each buffer, instead
+                # of letting each buffer's nested rmcb_list serializer issue its own query
+                rmcb_queryset = RecordMaxCounterBuffer.objects.filter(
+                    transfer_session_id=transfer_session_id,
+                    model_uuid__in={b.model_uuid for b in buffers},
```
rtibblesbot: **suggestion:** `{b.model_uuid for b in buffers}` is rebuilt from *all* buffers on every iteration of the `transfer_session_id` loop, not scoped to the current session. The comment above says the loop "limits the quantity of SQLite variables in use," but as written each iteration still passes the full unscoped set — the loop doesn't achieve that. Results stay correct today only because `BufferViewSet` always supplies a single transfer session, but it's a latent trap if a future caller batches multiple sessions. Consider scoping: `{b.model_uuid for b in buffers if b.transfer_session_id == transfer_session_id}`.

morango/api/serializers.py
```diff
@@ -158,11 +160,38 @@ class Meta:
         read_only_fields = fields
 
 
+class BufferListSerializer(serializers.ListSerializer):
```
rtibblesbot, line 163: **suggestion:** No test asserts query count for this `many=True` batching path — the repo has `test_buffer_serializer_makes_no_transfer_session_query` (test_api.py:1067) for the single-instance case, but nothing analogous pins down that serializing a batch of `Buffer`s now issues O(1)/O(#sessions) `RecordMaxCounterBuffer` queries instead of O(#buffers). Since eliminating that N+1 is the whole point of this PR, a `CaptureQueriesContext`-based test (mirroring the existing pattern) would guard against a future regression, e.g. someone dropping the `_rmcb_list` cache check in `Buffer.rmcb_list()`.

rtibblesbot (commented):
PR #345 review. Clean, minimal N+1 fix for `BufferListSerializer` — bulk-fetches `RecordMaxCounterBuffer` rows once per transfer session and caches them on each `Buffer` instance via `_rmcb_list`, with a backward-compatible fallback in `Buffer.rmcb_list()` for callers outside the list path. Correctness of the `(transfer_session_id, model_uuid)` bucketing checks out.

Note: new commits landed after this review's passes were run, so the latest changes on the branch have not been re-reviewed — please re-request review after this if anything material changed since. CI also shows failures/pending checks that should resolve before merge.

**Findings:**
- suggestion: `model_uuid__in` set isn't scoped to the current `transfer_session_id` inside the per-session loop, undercutting the comment's stated rationale (see inline).
- suggestion: no test locks in the new batched-query behavior for the `many=True` path (see inline).

---

</details>

rtibbles (approved):
Code updates make sense, and code comments explain why this _isn't_ resulting in a large number of queries still, as this will only usually operate over a single transfer session.

Existing test coverage gives further confidence of no regression.

morango/api/serializers.py
bjester: Fixed

morango/api/serializers.py
bjester: Fixed

morango/api/serializers.py
```diff
@@ -158,11 +160,38 @@ class Meta:
...
+            # Morango implementation will only ever call this for one transfer session ID at a time,
+            # but a loop makes the code straightforward regardless and limits the quantity of
+            # SQLite variables in use
+            for transfer_session_id in transfer_session_ids:
+                # bulk-fetch all RMCB records needed for this batch of buffers in a
+                # single query, then cache the relevant subset on each buffer, instead
+                # of letting each buffer's nested rmcb_list serializer issue its own query
+                rmcb_queryset = RecordMaxCounterBuffer.objects.filter(
+                    transfer_session_id=transfer_session_id,
+                    model_uuid__in={b.model_uuid for b in buffers if b.transfer_session_id == transfer_session_id},
```
rtibblesbot, line 178: **important:** This `IN` clause is sized to the number of buffers in the current page, and page size is driven by the caller-configurable sync `chunk_size` (`NetworkSyncConnection.chunk_size`, default 500 but overridable via `create_sync_session`/`resume_sync_session`). Since the codebase's only tested backend is SQLite, a caller that sets `chunk_size` above SQLite's compiled `SQLITE_MAX_VARIABLE_NUMBER` (999 on many stock builds) would hit `OperationalError: too many SQL variables` here — a failure mode that didn't exist when each buffer issued its own single-value filter. Consider sub-chunking the `IN` clause independent of `chunk_size`, or documenting/enforcing a max safe `chunk_size`.

morango/api/serializers.py
```diff
@@ -158,11 +160,38 @@ class Meta:
...
+class BufferListSerializer(serializers.ListSerializer):
+    def to_representation(self, data):
+        buffers = list(data)
+        if buffers:
+            rmcb_map = defaultdict(list)
+            transfer_session_ids = {b.transfer_session_id for b in buffers}
+            # Morango implementation will only ever call this for one transfer session ID at a time,
+            # but a loop makes the code straightforward regardless and limits the quantity of
+            # SQLite variables in use
+            for transfer_session_id in transfer_session_ids:
```
rtibblesbot, line 172: **suggestion:** This loop is written to handle multiple `transfer_session_id`s in one call (per the comment above), which is exactly the shape where the previously-fixed scoping bug lived. The new regression test only covers a single transfer session, so a future scoping regression on this branch wouldn't be caught. If this path is kept as future-proofing, add a multi-session test; otherwise consider simplifying to the single-session case that's actually exercised.

rtibblesbot (commented):
PR #345 delta re-review.

2 of 2 prior findings RESOLVED. No new correctness issues from the core pass. Two unresolved items from this round below; CI also still has several checks queued/pending as of this review.

- **important**: `model_uuid__in` set size scales with page size, which is driven by the caller-configurable sync `chunk_size` — see inline comment.
- **suggestion**: the multi-`transfer_session_id` branch of the new loop has no test coverage — see inline comment.

RESOLVED — morango/api/serializers.py:178 — model_uuid set not scoped to current transfer_session_id
RESOLVED — morango/api/serializers.py (test_api.py:1076) — missing test for batched-query behavior on `many=True` path

</details>

---

</details>

PR #346, opened 2026-07-08
Scoped field updates and context initialization issue

bjester:
## Summary
- Scopes model saves to `TransferSession` and `SyncSession` to only the fields that should have changed
- Adds regression test and fixes context initialization issue that could result in `is_push` of `None` when `TransferSession.push` is `False`

## TODO

## Reviewer guidance

The tests passed with the scoped changes, and the regression test covers the context bug. Is that sufficient?

The context bug is more important in the context of Kolibri's sync-extras-plugin, since it was passing in a hardcoded value to `is_push`

Kolibri's morango integration tests pass with these updates.

## Issues addressed

Related to this investigation:
- https://github.com/learningequality/kolibri-sync-extras-plugin/issues/14
- https://github.com/learningequality/kolibri-sync-extras-plugin/pull/15

##  AI Usage

AI found the issue in the `SessionContext`.

### DIFF
--- a/morango/sync/utils.py
+++ b/morango/sync/utils.py
@@ -137,16 +137,17 @@ def validate_and_create_buffer_data(  # noqa: C901
             model_uuid__in=[record["model_uuid"] for record in data]
         ).delete()
 
+        update_fields = ["records_transferred"]
         if connection is not None:
             transfer_session.bytes_sent = connection.bytes_sent
-        if connection is not None:
             transfer_session.bytes_received = connection.bytes_received
+            update_fields.extend(["bytes_sent", "bytes_received"])
 
         Buffer.objects.bulk_create(buffer_list)
         RecordMaxCounterBuffer.objects.bulk_create(rmcb_list)
 
         transfer_session.records_transferred += len(buffer_list) - deleted_buffers
-        transfer_session.save()
+        transfer_session.save(update_fields=update_fields)
 
 
 class SyncSignal(object):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ List of the most important changes for each release.
 
 ## 0.8.15
 - Improves buffer serialization performance by performing bulk counter lookup
+- Corrects an initialization issue with `SessionContext` that leads to an incorrect value for `is_push`
+- Save calls to Morango models during a sync are now scoped to only the changed fields
 
 ## 0.8.14
 - Adds utility for addressing immediate FK constraints caused by Django upgrade, automatically performed for morango models in a Django migration.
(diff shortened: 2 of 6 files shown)


morango/sync/utils.py
```diff
@@ -146,7 +146,7 @@ def validate_and_create_buffer_data(  # noqa: C901
         RecordMaxCounterBuffer.objects.bulk_create(rmcb_list)
 
         transfer_session.records_transferred += len(buffer_list) - deleted_buffers
-        transfer_session.save()
+        transfer_session.save(update_fields=["records_transferred"])
```
rtibblesbot: **blocking:** Lines 140-143 assign `transfer_session.bytes_sent`/`bytes_received` from the connection when `connection is not None`, but this save now only lists `records_transferred`, so those two columns are no longer persisted locally. The prior full `save()` wrote them. The sole caller passing a connection is `NetworkPullTransferOperation.handle` (operations.py:1684), so on a pull the local `bytes_sent`/`bytes_received` are dropped — the in-memory values are still read at operations.py:1699-1700 to update the *remote* session, so sync isn't broken, but local byte accounting is lost.

Build `update_fields` conditionally on whether `connection` was provided, e.g. `fields = ["records_transferred"]; if connection is not None: fields += ["bytes_sent", "bytes_received"]` then `save(update_fields=fields)`.

rtibblesbot (changes requested):
PR #346 scopes sync-stage `save()` calls to `update_fields` and fixes the `SessionContext.is_push` coercion bug. The context fix and its regression test are clean, but one `update_fields` list drops fields the same code path assigns.

- **blocking** — `validate_and_create_buffer_data` no longer persists `bytes_sent`/`bytes_received` locally (inline).
- nitpick — a few `update_fields` lists include fields the branch doesn't always mutate (`operations.py:1112` `server_fsic` on the client branch; `operations.py:1504` `records_total` on push; `core.py:336` both stage fields). Harmless since in-memory equals persisted, but if the goal is strict scoping they could be narrowed. Not worth changing on their own.
- suggestion — no test pins the "only these fields are written" contract; a future edit that mutates a field but forgets to extend `update_fields` would silently regress with no failure. The `is_push` fix is well-tested; consider a targeted test that mutates a field out-of-band and asserts a scoped save leaves it intact.

CI: the plain unit-test matrix was pending and the app-token check is failing (unrelated permissions step, not this diff) — hence COMMENT-level CI, but the byte-accounting bug drives REQUEST_CHANGES.

---

</details>

morango/sync/utils.py
bjester: fixed

rtibbles (approved):
Changes make sense and align with the PR description.

PR #348, opened 2026-07-28
Merge v0.8.x branch into v0.9.x

bjester:
## Summary
Merges the branches and resolves merge conflicts:
```
	both modified:   morango/__init__.py
	both modified:   morango/sync/session.py
	both modified:   morango/sync/syncsession.py
	deleted by us:   setup.py
	both modified:   tests/testapp/tests/sync/test_session.py
	both modified:   tests/testapp/tests/test_api.py
```

Also updated the migration history for conflicting migration nodes


morango/migrations/0005_self_ref_order.py
```diff
@@ -7,7 +7,7 @@
 
 class Migration(migrations.Migration):
     dependencies = [
-        ("morango", "0003_store_deserialization_errors"),
+        ("morango", "0004_deferrable_foreign_keys"),
```
rtibblesbot, line 10: **blocking:** Renaming this migration silently wipes `_self_ref_order` on existing 0.9.x databases.

Keeping the released `0004_deferrable_foreign_keys` name is right (it shipped in v0.8.14), but Django keys applied migrations on `(app, name)` — so a database that already recorded `morango.0004_self_ref_order` sees `0005_self_ref_order` as unapplied and re-runs it. It doesn't error: Django 3.2's SQLite backend implements this `AddField` as a table rebuild whose `INSERT ... SELECT` supplies a literal `NULL` for the "new" column.

Reproduced locally — migrated `upstream/release-v0.9.x` onto a fresh SQLite DB, seeded a `morango_store` row with `_self_ref_order = 7`, then migrated this branch over it:

```
Applying morango.0004_deferrable_foreign_keys... OK
Applying morango.0005_self_ref_order... OK
AFTER: [('0', None)]     # was 7
```

`Store._self_ref_order` is persisted ordering consumed by `morango/sync/stream/serialize.py` and the `_queue_into_buffer` SQL, so this is real data loss. Scope is limited to pre-release databases tracking `release-v0.9.x` (Kolibri `develop`, CI caches) since there are no `v0.9*` tags — but it fails silently, which makes it easy to miss.

**Suggested fix — put the incoming migration last instead of renumbering this one.** I verified this works:

- keep this file as `0004_self_ref_order`, dependency `("morango", "0003_store_deserialization_errors")`
- rename `0004_deferrable_foreign_keys` → `0005_deferrable_foreign_keys`, dependency `("morango", "0004_self_ref_order")`

```
Applying morango.0005_deferrable_foreign_keys... OK
AFTER: [('0', 7)]        # preserved
```

For 0.8.14/0.8.15 users the recorded `0004_deferrable_foreign_keys` simply drops out of the graph and the renamed node re-runs `MakeForeignKeysDeferrable` — which is idempotent. I confirmed that too: after the rebuild, `_table_has_immediate_foreign_key` returns `False` for `morango_store`, `morango_buffer` and `morango_recordmaxcounter`, so the re-run is a no-op and touches no data.

Two alternatives I tested and would avoid:

- `replaces = [("morango", "0004_self_ref_order")]` on this file preserves the data, but leaves `0004_deferrable_foreign_keys` permanently unapplied (`showmigrations` shows `[ ]`, and `migrate` reports "No migrations to apply") — the FK fix would silently never run on those databases.
- Keeping both at `0004` and depending on `0004_deferrable_foreign_keys` raises `InconsistentMigrationHistory` on existing 0.9.x DBs.

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,137 @@
...
+                logger.info(
+                    "Rebuilding table %s to make foreign key constraints deferrable",
+                    table_name,
+                )
+                try:
+                    # Rebuilding the table from the current model state regenerates the column
+                    # definitions, which SQLite always emits with the deferrable clause (see
+                    # DatabaseSchemaEditor.sql_create_inline_fk). Only tested with Django 3.2
+                    schema_editor._remake_table(model)
+                except DatabaseError as e:
```
rtibblesbot, line 126: **suggestion:** Swallowing `DatabaseError` here means the migration is still recorded as applied, so it will never be retried — the immediate-FK constraints this module exists to fix stay broken permanently, with only a log line most Kolibri installs will never surface.

It's also risky mid-rebuild: `_remake_table` is `CREATE new__<table>` → `INSERT … SELECT` → `DROP` → `RENAME`, and SQLite doesn't abort the enclosing transaction on a failed statement. A failure after the `DROP` lets the loop continue and the migration's atomic block commit the partial state.

Letting the error propagate (or collecting failures and re-raising after the loop) would fail loudly and stay re-runnable. If a soft-fail is genuinely wanted for the manual `run()` entry point, an opt-in constructor flag would keep it out of the migration path.

morango/deferrable_foreign_keys.py
```diff
@@ -0,0 +1,137 @@
...
+    clause without ``DEFERRABLE`` indicates an immediate constraint that needs to be rebuilt.
+    """
+    sql = _get_table_sql(conn, table_name)
+    if not sql:
+        return False
+    references_count = sql.count("REFERENCES")
+    if references_count == 0:
+        return False
+    # make sure that all FK columns (identified by references) have a deferrable constraint
+    return references_count != sql.count("DEFERRABLE")
```
rtibblesbot, line 66: **suggestion:** Substring-counting the DDL is fragile — any column name, `CHECK` expression, or string `DEFAULT` containing `REFERENCES` or `DEFERRABLE` skews the comparison, causing either a pointless full-table rebuild or a table that never gets fixed. No morango or Kolibri table hits this today, so this is robustness rather than a live bug; `PRAGMA foreign_key_list(<table>)` cross-checked against the `sqlite_master` text would be sturdier.

Separately, `tests/testapp/tests/test_deferrable_foreign_keys.py` covers `__call__` and `_iter_apps` but not `run()` — the entry point downstream consumers are meant to call, and the one with a real precondition (SQLite's `disable_constraint_checking()` returns `False` inside an open transaction, making `schema_editor.__enter__` raise `NotSupportedError`). Worth a test that `run()` works outside a transaction.

morango/api/serializers.py
```diff
@@ -157,11 +159,40 @@ class Meta:
         read_only_fields = fields
 
 
+class BufferListSerializer(serializers.ListSerializer):
+    def to_representation(self, data):
+        buffers = list(data)
+        if buffers:
+            rmcb_map = defaultdict(list)
+            transfer_session_ids = {b.transfer_session_id for b in buffers}
+            # Morango implementation will only ever call this for one transfer session ID at a time,
```
rtibblesbot, line 168: **suggestion:** The comment claims the loop limits SQLite variables, but it doesn't — the loop is over transfer session ids (always one, as the comment itself says), while the `model_uuid__in` parameter count scales with the number of buffers in the batch.

Morango's own client always sends `limit=self.chunk_size` (default 500), so this is fine in practice. But `BufferViewSet` uses `LimitOffsetPagination` with no `max_limit` and DRF's `PAGE_SIZE` is unset, so a `GET /buffers/?transfer_session_id=…` with no `limit` disables pagination and produces one placeholder per buffer in the session — `too many SQL variables` on SQLite < 3.32. Previously that request was an N+1: slow, but it completed.

Same unbounded shape on the push-receive path at `morango/sync/utils.py:130` and `:134`, sized by the client-supplied POST body. Chunking the uuid list in both places (and/or setting `max_limit` on the pagination class) would remove the ceiling. Worth fixing the misleading comment either way.

morango/sync/session.py
```diff
@@ -44,56 +122,195 @@ def __init__(self):
...
+            return None
+
+        # requests might wrap `MaxRetryError` in a `ConnectionError`
+        if any(isinstance(arg, MaxRetryError) for arg in e.args):
+            return None
+
+        try:
+            # may raise if retries have been exhausted
+            retries = retries.increment(
+                method=request.method, url=request.url, response=response, error=e
```
rtibblesbot, line 306: **suggestion:** `response` is always `None` at this point, and passing it is the risky part of the plumbing.

It's reset to `None` at the top of each loop iteration and only assigned on the success path, so the `except` branch can only ever pass `response=None` — `_should_retry` then falls through to `response = e.response` anyway. That makes the parameter dead.

It's also a duck-typing trap: `Retry.increment` and `retries.sleep` expect a `urllib3.response.HTTPResponse` and call `get_redirect_location()`, `.status`, and (on urllib3 1.26, permitted by the new floor) `.getheader("Retry-After")`. A `requests.Response` has none of those. Today the truthiness guards short-circuit because all three members of `_RETRY_REQUEST_EXCEPTIONS` are raised without a `response` attached — but any future change that surfaces a retryable `RequestException` carrying one turns a would-be retry into `AttributeError`, escaping the `except MaxRetryError` handler as an unrelated error type.

Passing `response=None` explicitly (the retry decision is already driven entirely by exception type) removes both the dead parameter and the trap.

morango/sync/session.py
```diff
@@ -44,56 +122,195 @@ def __init__(self):
...
+        else:
+            ctx = nullcontext()
+
+        with ctx:
+            response = None
+            try:
+                response = super(SessionWrapper, self).request(method, url, **kwargs)
+                response.raise_for_status()
+                return response
+            except Exception as e:
```
rtibblesbot, line 237: **suggestion:** Widening from `except exceptions.RequestException` to `except Exception` routes genuine programming errors — a `TypeError` from a bad kwarg, an `AttributeError` from a mock — through `_log_response_error`, which logs them as `TypeError Reason: (no response)`. That reads like a network failure and buries the traceback.

`raise_for_status()` raises `HTTPError`, itself a `RequestException` subclass, so narrowing back loses no coverage.

morango/sync/session.py
```diff
@@ -23,19 +42,78 @@ def _headers_content_length(headers):
...
 class SessionWrapper(Session):
     """
     Wrapper around `requests.sessions.Session` in order to implement logging around all request errors.
     """
 
-    bytes_sent = 0
-    bytes_received = 0
-
-    def __init__(self):
+    def __init__(self, max_retries):
```
rtibblesbot, line 109: **suggestion:** `max_retries` is now a required positional argument on a class that's importable public API (`morango.sync.session.SessionWrapper`) and was previously constructible with no arguments — any external caller doing `SessionWrapper()` breaks on upgrade.

Defaulting to `max_retries=None` and falling back to the `HTTPAdapter` default (`Retry(0, read=False)`, i.e. today's non-retryable behaviour) keeps the new capability opt-in at no cost, since retries are already opt-in per request.

morango/sync/syncsession.py
```diff
@@ -132,15 +152,19 @@ def __init__(
...
         # with 7 retry attempts, sleep escalation becomes (0.6s, 1.2s, ..., 38.4s)
-        retry = Retry(total=retries, backoff_factor=backoff_factor)
-        adapter = HTTPAdapter(max_retries=retry)
-        self.session.mount("http://", adapter)
-        self.session.mount("https://", adapter)
+        self.session = SessionWrapper(
+            Retry(
+                total=retries,
+                backoff_factor=backoff_factor,
+                allowed_methods=None,  # allow any method
```
rtibblesbot, line 161: **suggestion:** `allowed_methods=None` makes POST/PATCH/DELETE retryable at the transport layer, where urllib3's default set previously excluded them. The idempotency work in this PR covers the cases that need it — `_push_record_chunk` via delete-then-recreate, the session-close DELETEs via `ignore_404`, and `_create_transfer_session` because the server reuses an existing active session. `_get_nonce` is harmless (an extra `Nonce` row).

Two things worth confirming rather than inferring:

- Is `_create_sync_session` (line 458) deliberately left non-retryable while `_create_transfer_session` is retryable? A dropped connection during sync-session creation is exactly the flakiness this targets, so the asymmetry reads as unintentional.
- The `send()` retry loop re-sends `request` as-is, which is fine for morango's own bodies (bytes/JSON, never streamed) but would silently send an empty body for a file-like or generator body. Worth a comment noting the assumption.

morango/models/core.py
```diff
@@ -844,6 +855,11 @@ def save(self, update_dirty_bit_to=True, *args, **kwargs):
             self._morango_dirty_bit = True
         elif not update_dirty_bit_to:
             self._morango_dirty_bit = False
+
+        # ensure the dirty bit field is in the fields to update if present, to keep it in sync
+        if update_dirty_bit_to is not None and kwargs.get("update_fields") is not None:
```
rtibblesbot, line 860: **nitpick:** Because `update_dirty_bit_to` takes the first positional slot, `force_insert`/`force_update`/`using`/`update_fields` land in `*args` when passed positionally and `kwargs.get("update_fields")` misses them — the in-memory `_morango_dirty_bit` would be set but never written. Django itself always uses keywords and both new tests use the keyword form, so this is unlikely to bite. Making the signature keyword-only would close it; nothing to do if the keyword contract is considered documented.

morango/sync/utils.py
```diff
@@ -125,18 +125,27 @@ def validate_and_create_buffer_data(  # noqa: C901
...
-        if connection is not None:
             transfer_session.bytes_received = connection.bytes_received
-
-        transfer_session.save()
+            update_fields.extend(["bytes_sent", "bytes_received"])
 
         Buffer.objects.bulk_create(buffer_list)
         RecordMaxCounterBuffer.objects.bulk_create(rmcb_list)
 
+        transfer_session.records_transferred += len(buffer_list) - deleted_buffers
```
rtibblesbot, line 146: **praise:** Adjusting by `len(buffer_list) - deleted_buffers` keeps `records_transferred` accurate across a re-pushed chunk rather than double-counting or resetting, and `test_push_repeat_chunk` pins exactly that behaviour.

rtibblesbot (changes requested):
PR #348 — merge of `release-v0.8.x` into `release-v0.9.x`. The merge is clean and the 0.8.12–0.8.15 work (idempotent buffer push, opt-in retries, `update_fields` scoping, dirty-bit fix) carries over faithfully.

One blocking issue, in the migration-history resolution: renaming `0004_self_ref_order` → `0005_self_ref_order` silently nulls out `Store._self_ref_order` on any database that already applied the old name. I reproduced it locally and verified a fix — details inline.

Remaining findings are suggestions around the new retry plumbing, the swallowed `DatabaseError` in the FK-deferrable migration, and the unbounded `model_uuid__in` queries.

CI: all unit tests (3.9–3.13, postgres, SQLite matrix) and linting pass. The one red check, `Check if author is contributor`, fails in the shared `call-pull-request-target.yml` workflow — repo/app-token infrastructure, not this diff.

No UI files changed, so visual verification and manual QA do not apply.

---

</details>

morango/migrations/0005_self_ref_order.py
bjester: You're confused. This migration has never shipped-- there is no release of morango v0.9 yet. This is a non issue.

rtibblesbot: Understood, deferring to @bjester — withdrawing this finding. Apologies for the noise.

morango/api/serializers.py
bjester: The comment doesn't claim that the loop itself limits the query variables.

rtibblesbot: You're right — read as "partitioning by session bounds each `IN` clause to that session's buffers rather than one clause over all of them," the loop does limit variables, and my reading of the comment was wrong. Withdrawing that part.

The unbounded `model_uuid__in` when a client omits `limit` stands as written in the original comment; no need to relitigate the wording.

rtibbles (approved):
All changes look of a piece with the changes in the release-v0.8.x branch - no concerns and tests are green.

